PR #29, opened 2025-08-25
Response to Issue #27: Clarify MUST ignore requirement for unknown parameters

### LINKED ISSUE #27: Ignoring unknown keys
The draft states that "CAs **MUST** ignore any unknown parameter keys." This seems like a bad idea.

I get the need to have an extensible protocol, but a future extension may have undesirable semantics when a certain field is ignored. What would be better is to be able to specify critical keys, that the CA must understand to proceed.

sheurich:
This PR addresses Issue #27 regarding the requirement that CAs MUST ignore any unknown parameter keys.

### 1. RFC 8555 Precedent
The ACME protocol (RFC 8555) establishes this pattern in multiple contexts, including:
- Unknown fields in objects (Section 6.2)
- Unknown challenge types
- Unknown identifiers
This creates a consistent pattern across ACME specifications.

### 2. Forward Compatibility
The MUST ignore requirement ensures:
- Future extensions can add new parameters without breaking existing implementations
- Graceful evolution of the protocol without version negotiation
- Smooth deployment of new features

### 3. Operational Simplicity
- Avoids the complexity of a critical key system
- Reduces implementation burden on CAs
- Prevents unnecessary failures for harmless unknown parameters
- Simplifies client-server interactions

## Changes
- The requirement to ignore unknown parameters is clarified to include the rationale (forward compatibility) and to specify "the server" as the actor.
- The term "parameter key(s)" has been replaced with the more precise term "tag(s)" throughout the document, consistent with RFC 8659.
- A direct reference to the `tag=value` syntax in RFC 8659 has been added to improve clarity at the point of first use.

Closes #27

### DIFF
--- a/draft-sheurich-acme-dns-persist.md
+++ b/draft-sheurich-acme-dns-persist.md
@@ -163,7 +163,11 @@ The RDATA of this TXT record MUST fulfill the following requirements:
 
 3.  The issue-value MUST contain an accounturi parameter. The value of this parameter MUST be a unique URI identifying the account of the applicant which requested the validation, constructed according to {{!RFC8657}}, Section 3.
 
-4.  The issue-value MAY contain a `policy` parameter. If present, this parameter modifies the validation scope. The `policy` parameter follows the `key=value` syntax. The policy parameter key and its defined values MUST be treated as case-insensitive. CAs MUST ignore any unknown parameter keys. The following value for the `policy` parameter is defined with respect to subdomain and wildcard validation:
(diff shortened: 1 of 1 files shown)


birgelee:
LGTM.

Also a point for this is that the syntax is nearly identical to the CAA issue tag which also upholds this proceed to issue in the presence of unknowns extensions behavior.

Being the author of (I think the first) critical bit CAA tag, I agree that the critical system hampers rapid innovation while adding significant complexity/confusion. If a new blocking security control is actually needed, it can be put as a critical CAA tag and should probably not be in the DCV flow anyways.

PR #30, opened 2025-10-07
feat: add accounturi attribute to the challenge object

slghtr-says:
Added accounturi attribute to the challenge response object.

Changes include:
- Updated accounturi references to refer to the server provided accounturi rather than "a valid account uri"
- Cleaned up accounturi references and examples.

**Motivation**: This change eliminates ambiguity and improves interoperability of implementations. While ACME clients are aware of their account URI, explicitly providing this value in the challenge object removes any dependency on client-side URI construction or format assumptions. This approach makes the exact value the server expects in DNS explicit and ensures the challenge object is completely self-contained and driven by the server.

### DIFF
--- a/draft-ietf-acme-dns-persist.md
+++ b/draft-ietf-acme-dns-persist.md
@@ -3,7 +3,7 @@ title: "Automated Certificate Management Environment (ACME) Challenge for Persis
 abbrev: "ACME Persistent DNS Challenge"
 category: std
 submissionType: IETF
-docname: draft-sheurich-acme-dns-persist-latest
+docname: draft-ietf-acme-dns-persist-latest
 ipr: trust200902
 area: "Security"
 workgroup: "Automated Certificate Management Environment"
@@ -27,8 +27,8 @@ author:
    email: "sheurich@fastly.com"
  -
    name: "Henry Birge-Lee"
-   org: "Princeton University"
-   email: "birgelee@princeton.edu"
+   org: "Crosslayer Labs, Inc."
+   email: "henry@crosslayerlabs.com"
  -
    name: "Michael Slaughter"
    org: "Amazon Trust Services"


birgelee:
Hi all,

I can probably clean up the build error on this.

That said, I was wondering if there is any other language that could improve specificity while also permitting there to possibly be multiple account uris values that permit issuance by a single account. The motivation is twofold:

1) in the future allow for orgs to obfuscate the relationship of multiple domains using the same ACME account. This could make that ACME account a target for hijack or allow adversaries to discover organizational infrastructure that they otherwise would not be able to pin back to the host organization.

2) allow more fine-grained access controls within an ACME account. I am not sure if any CAs have adopted this model, but particularly when EAB (external account binding) is used or rate limit exceptions are in play, there may be a desire for a single ACME account to be subdivided into different policy roles each one authorized for some subset of domains. In this nature, an account id could be used to refer to a specific policy role and apposed to mapping to a true ACME account. I can't tell if this really makes sense (as these roles need independent credentials, so they are sort of just ACME accounts themselves), but I sort of like the idea of this field being a more abstract notion of identity than a forced 1-to-1 binding with an ACME account.

birgelee:
tl:dr; I am going to keep this open for discussion and merge a procedural posh into main to change the draft name. This can be fast forwarded and then merged in.

I am doing a push to change the draft name (post adoption it is draft-ietf-acme-dns-persist) and also update my affiliation (I am not longer a researcher at Princeton University but a CEO of Crosslayer Labs).

I was tempted to merge this in first, but I would still like to discuss it a bit. My understanding is that with this new system the CA must send the accounturi first and then the client MUST use that exact accounturi. However, if there are multiple potentially valid values for an accounturi, I don't know how the CA predicts which one to send.

Perhaps there is a more fundamental conversation going on here: is the client actually provisioning this record or is it just action under the authority of another entity that already provisioned the record prior to the client's new-order request?

I somehow always imagine the later case in which scenario less on the wire ensures maximum flexibility for the CA to determine the validity of the accounturi value independently. If the client is expected to dynamically provision this record, I do see how having the CA send the accounturi reduces load on the client and eliminates the need for the client to have self awareness of the accounturi (although I believe most ACME clients do know their accounturi).

Perhaps a middle ground could be to have the CA send one accounturi which the client knows it could use but put in some language about the CA possibly accepting other accounturi values so long as they appropriately authorize the requesting ACME account.

I think this would benefit from discussion on a call. Also, @sheurich do you have any thoughts from the client side on this?

PR #35, opened 2026-02-19
decouple accounturi validation from challenge object

sheurich:
Builds on #30 by @slghtr-says. Rebased onto main with additional changes by @sheurich.

### Summary

PR #30 added `accounturi` to the challenge object. The changes here address a coupling issue: #30 defined `accounturi` validation in terms of the challenge object, but the challenge object does not exist in the JIT validation path or when records are pre-provisioned.

### Changes

**Decouple `accounturi` validation from the challenge object.** The DNS record requirement (§4 item 3) now stands on its own: the `accounturi` MUST be a URI identifying the requesting ACME account, and the CA MUST verify it. No reference to the challenge object. Works for JIT, pre-provisioned, and standard challenge flows.

**Many-to-one `accounturi` model.** CAs MUST accept the ACME account URL (RFC 8555 §7.3). CAs MAY also accept alternative URIs, provided each uniquely and permanently identifies a single account (RFC 8657 §5.4). Non-reassignment and deactivation lifecycle requirements included. This addresses the privacy concern raised by @birgelee — publicly visible `accounturi` values in long-lived DNS records create a correlation channel (RFC 8657 §5.9). See the new Account URI Privacy subsection in Security Considerations.

**`accounturi` in the challenge object remains required.** The field is the mechanism by which the CA communicates the expected URI (including alternative URIs) to the client.

**Add `accounturi` comparison semantics.** Simple String Comparison per RFC 3986 §6.2.1. PR #36 adds a consolidated comparison paragraph for all parameter values; the inline rule here serves as a standalone fallback.

**Add pre-provisioning guidance (new §4.4).** How to construct records without a challenge object: use the ACME account URL from the `Location` header, obtain `issuer-domain-name` from `caaIdentities` as a hint (not normatively coupled). Operational guidance: maintain record inventory, use `persistUntil`, audit after DNS incidents. CAs SHOULD maintain stable account URIs and document their guarantees.

**Add JIT rate-limiting and abuse guidance.** Per-domain rate limiting independent of requesting account. CAs SHOULD restrict JIT to accounts with prior successful `dns-persist-01` validation (trust-on-first-use). Account activity logging for the JIT path.

**Fix DNSSEC failure semantics.** Resolved a SHOULD/MUST inconsistency: DNSSEC validation remains SHOULD, but if a CA performs it, failure MUST be treated as a challenge failure. Includes rationale for the stricter-than-RFC-8555 requirement.

**Fix DNS caching caveat.** Revocation section previously said "after the record is removed, new validation attempts will fail" — misleading. Now says "after the record is removed and resolver caches have expired."

**Neutralize §4 example prose.** Changed "it might provision" to "the DNS TXT record would contain" — removes assumption that the client provisions after receiving the challenge. Add `accounturi` to §Examples challenge objects for consistency.

**Error handling.** Changed "does not match the URI" to "does not identify the ACME account" — covers both account URL comparison and alternative URI resolution.

**Terminology.** Use "URI" consistently (matching the parameter name and RFC 8657), "ACME account URL" only for the RFC 8555 §7.3 concept.


draft-ietf-acme-dns-persist.md
```diff
@@ -167,7 +169,7 @@ The RDATA of this TXT record MUST fulfill the following requirements:
 
 2.  The `issuer-domain-name` portion of the issue-value MUST be one of the Issuer Domain Names provided by the CA in the `issuer-domain-names` array of the challenge object.
 
-3.  The issue-value MUST contain an accounturi parameter. The value of this parameter MUST be a unique URI identifying the account of the applicant which requested the validation, constructed according to {{!RFC8657}}, Section 3.
+3.  The issue-value MUST contain an `accounturi` parameter. The value of this parameter MUST be the ACME account URL of the account requesting validation, as defined in {{!RFC8657}}, Section 3. The CA MUST verify that this URI identifies the ACME account making the request. When comparing `accounturi` values, implementations MUST use simple string comparison as defined in {{!RFC3986}}, Section 6.2.1, and MUST NOT perform URI normalization, percent-decoding, or case-folding before comparison.
```
birgelee: I have some serious concerns with tight coupling this to the Account URI using a 1-to-1 mapping. Right now, with the exception of RFC8657 users (which is a very small percentage of domains) there is no way from public logs to correlate two domains back to the same ACME account. Being able to do so opens up tons of infrastructure discovery avenues as two domains using the same ACME account indicates with high probability those domains are hosted on the same server/hosting infrastructure. Then, if there is a vulnerability in the software stack at one domain, it could be used to compromise the other domain.

I am hoping to have this method see widespread adoption and I think if it leaks ACME account bindings it could hinder adoption.

Nothing in Michael Slaughter's CAB/F ballot forces this type of coupling. I have already spoken with several CA owners about plans to either using hashing or a back-end database maintained by the CA to issue virtual URIs for the purpose of DCV.

I think the main security parameter needed for this uri is that there exists some function F known to the CA such that F(URI) -> exactly 1 ACME account in perpetuity.  There can be multiple URIs that produce the same ACME account.

Additionally a small change: I am not aware of any definition of acme account URL in RFC8657. Should "the ACME account URL of the account" be changed to "ACME account URI"?

I also, I know it was in the original draft, but I don't super love RFC8657 as a reference since it is not the ACME RFC itself and RFC8657 is generic to ACME and non-ACME while this I-D is ACME specific. I feel like this is a bit of an indirect reference. Sec 3.1 of RFC 8657 defines "URI of the ACME account object" when making the proposal ACME specific.

draft-ietf-acme-dns-persist.md
```diff
@@ -148,6 +136,7 @@ The following shows an example challenge object:
   "type": "dns-persist-01",
   "url": "https://ca.example/acme/authz/1234/0",
   "status": "pending",
+  "accounturi": "https://ca.example/acct/123",
```
aarongable: optional nit, to match the format of the url above?
```suggestion
  "accounturi": "https://ca.example/acme/acct/123",
```

Issue #34, opened 2026-02-13
Specific case-Insensitivity carveout for the value of the policy tag


beautifulentropy:
While implementing dns-persist-01 support in [Pebble](https://github.com/letsencrypt/pebble/pull/536), I noticed an inconsistency in how case-sensitivity is defined for issue-value parameters.

```
The issue-value MAY contain a policy parameter. If present, this parameter modifies the validation scope. The policy parameter follows the ‘tag=value’ syntax from {{!RFC8659}}. The parameter’s ‘tag’ and its defined values MUST be treated as case-insensitive.
```

RFC 8659 specifies that tags are compared case-insensitively, but it does not define case-handling rules for tag values. The dns-persist-01 draft, however, seems to explicitly require the value of the policy tag (for example, policy=wildcard) to be treated as case-insensitive.

The same requirement does not appear to apply to the values of the other tags defined by the draft. From an implementation perspective this raises a question: What is special about the value of the policy tag that warrants case-insensitive handling, while other tag values remain unspecified?

PR #36, opened 2026-02-19
Clarify case-handling rules for parameter values

sheurich:
Addresses #34.

## Problem

The draft requires case-insensitive comparison for `policy` values but says nothing about the case-handling of `accounturi` or `persistUntil` values. An implementer (@beautifulentropy, working on [Pebble support](https://github.com/letsencrypt/pebble/pull/536)) asked: what makes `policy` special?

The asymmetry is correct, but the draft doesn't explain it:
- `accounturi` values are URIs — case-sensitive per RFC 3986
- `persistUntil` values are integers — case is inapplicable
- `policy` values are protocol keywords — case-insensitivity is a deliberate usability choice

RFC 8659 §4.1 makes parameter *tags* case-insensitive but defines no rules for parameter *values*, leaving that to whichever specification defines the parameter.

## Changes

**Item 4 (policy parameter):** Replace the ambiguous sentence *"The parameter's 'tag' and its defined values MUST be treated as case-insensitive"* with a statement scoped to `policy` values only. The tag clause was redundant (already covered by RFC 8659 §4.1).

**New paragraph after item 5:** State the case-handling rules for all parameter values in one place:
- `accounturi`: Simple String Comparison per RFC 3986 §6.2.1; MUST NOT case-fold
- `policy`: case-insensitive (protocol keywords, often manually provisioned)
- `persistUntil`: base-10 integer; case does not apply

## Compatibility with #35

PR #35 adds `accounturi` comparison semantics to item 3 (simple string comparison, no normalization/case-folding). The new paragraph here reinforces that without conflicting. Both PRs can merge in either order.

### DIFF
--- a/draft-ietf-acme-dns-persist.md
+++ b/draft-ietf-acme-dns-persist.md
@@ -178,7 +178,7 @@ The RDATA of this TXT record MUST fulfill the following requirements:
 
 This specification defines the following case-handling rules for parameter values in dns-persist-01 records:
 
-- `accounturi`: The value is a URI. Implementations MUST compare `accounturi` values using Simple String Comparison per {{!RFC3986}}, Section 6.2.1, with no case-folding or other normalization.
+- `accounturi`: The value is a URI. CAs MUST compare `accounturi` values using Simple String Comparison per {{!RFC3986}}, Section 6.2.1, with no case-folding or other normalization.
 - `policy`: Case-insensitive, as specified in item 4 above.
 - `persistUntil`: The value is a base-10 integer. Case does not apply.
 


draft-ietf-acme-dns-persist.md
```diff
@@ -181,6 +181,12 @@ The RDATA of this TXT record MUST fulfill the following requirements:
 
 5.  The issue-value MAY contain a `persistUntil` parameter. If present, the value MUST be a base-10 encoded integer representing a UNIX timestamp (the number of seconds since 1970-01-01T00:00:00Z ignoring leap seconds). CAs MUST NOT consider this validation record valid for new validation attempts after the specified timestamp. However, this does not affect the reuse of already-validated data.
 
+{{!RFC8659}}, Section 4.1, requires parameter tags to be matched case-insensitively but does not define case-handling rules for parameter values. This specification defines the following rules:
```
birgelee: minor comment: in my experience its usually not good to say (again) what another standard "does."  I think there is some benefit in explaining this but the first sentence with the RFC 8659 reference also introduces some risk. If somehow there is errata or a new RFC comes along that contradicts this, this line could create confusion. Unless its needed we may just want to start with " we define the following case rules for case processing in dns persist 01 records" or something along those lines.

birgelee:
Nice work. I think this helps the document. I had a nit but nothing really major with this.

sheurich:
Pushed f501825 addressing @birgelee's review plus three additional cleanups:

- Opening sentence: Removed RFC 8659 §4.1 paraphrase per review feedback. Now states rules directly.
- accounturi bullet: Removed redundant MUST NOT case-fold (entailed by Simple String Comparison) and incorrect §6.2.2.1 citation. Replaced with "with no case-folding or other normalization" as a subordinate clause.
- policy bullet: Now cross-references item 4 instead of restating the MUST. Consolidated list is a quick-reference, not a second normative source.
- Example sentence: Adopted PR #35's rewording to reduce merge conflict surface. Whichever PR merges second will see a one-side insertion conflict (the case-handling block) rather than a text conflict on two overlapping edits.

birgelee:
This looks great. @sheurich feel free to merge in as fit.

Issue #33, opened 2026-01-12
Subdomain validation prompting


aarongable:
Subdomain Validation is a cool concept, and one that I'm happy dns-persist is supporting. In the terms of the Baseline Requirements, this is equivalent to selecting an Authorization Domain Name by pruning labels from the left-hand side of the FQDN, and then performing validation at that ADN.

However, ACME does not have good support for selecting ADNs. Traditional CAs, or CAs which use External Account Binding, might have ways for Applicants to indicate to the Server that they want to use a particular ADN for all of their validation. But plain ACME CAs without EAB would have no way to know whether the Applicant wants to validate at `server.dept.example.com`, at `dept.example.com`, or at `example.com`. I'm not a fan of the server having to fan out a bunch of prospective queries to see what layer of the DNS hierarchy the Applicant has placed their dns-persist record at.

So it would be great if there was a way for the Applicant to proactively indicate to the CA that they want to take advantage of Subdomain Validation. My proposal would be that the ACME client can include an ADN in their POST request to the Challenge URL. The CA would then check that the supplied URL is a valid ADN for the Authorization, and if it is, perform validation at that URL instead. And of course, as already described by the document, ensure that `policy=wildcard` is set.

-----

i.e. if we have an authorization that looks like this:
```
   HTTP/1.1 200 OK
   Content-Type: application/json
   Link: <https://acme.ca/directory>;rel="index"

   {
     "status": "pending",
     "expires": "2026-02-01T14:09:30Z",

     "identifier": {
       "type": "dns",
       "value": "admin.blog.example.com"
     },

     "challenges": [
       {
         "type": "http-01",
         "url": "https://acme.ca/chall/PAniVnsZcis/http-01",
         "token": "DGyRejmCefe7v4NfDGDKfA"
       },
       {
         "type": "tls-alpn-01",
         "url": "https://acme.ca/chall/PAniVnsZcis/tls-alpn-01",
         "token": "DGyRejmCefe7v4NfDGDKfA"
       },
       {
         "type": "dns-01",
         "url": "https://acme.ca/chall/PAniVnsZcis/dns-01",
         "token": "DGyRejmCefe7v4NfDGDKfA"
       },
       {
         "type": "dns-persist-01",
         "url": "https://acme.ca/chall/PAniVnsZcis/dns-persist-01",
         "token": "DGyRejmCefe7v4NfDGDKfA"
       },
     ]
   }
```

[...]

PR #41, opened 2026-03-04
fix(terminology): rename "Authorization Domain Name" to "Validation Domain Name"

### LINKED ISSUE #40: Use of "Authorization Domain Name" defined term
This draft defines "Authorization Domain Name (ADN)" as:

> The domain name at which the validation TXT record is provisioned. It is formed by prepending the DNS TXT Record Persistent DCV Domain Label to the FQDN being validated.

In contrast, the Baseline Requirements define "Authorization Domain Name (ADN)" as:

> The FQDN used to obtain authorization for a given FQDN to be included in a Certificate. The CA may use the FQDN returned from a DNS CNAME lookup as the FQDN for the purposes of domain validation. If a Wildcard Domain Name is to be included in a Certificate, then the CA MUST remove "*." from the left-most portion of the Wildcard Domain Name to yield the corresponding FQDN. The CA may prune zero or more Domain Labels of the FQDN from left to right until encountering a Base Domain Name and may use any one of the values that were yielded by pruning (including the Base Domain Name itself) for the purpose of domain validation.

Accompanying this definition, Section 3.2.2.4.22 DNS TXT Record with Persistent Value says:

> The record MUST be placed at the "_validation-persist" label prepended to the Authorization Domain Name being validated (i.e., "_validation-persist.[Authorization Domain Name]").

In other words, per the BRs, `example.com` might be the ADN, while `_validation-persist.example.com` is a nameless thing scoped solely to Section 3.2.2.4.22. The fact that this draft calls `_validation-persist.example.com` the ADN leads to confusion when reading and implementing this spec, and we should probably use a different term.

sheurich:
Closes #40

## Problem

The draft defines "Authorization Domain Name" as the full DNS label where the validation TXT record is provisioned (e.g., `_validation-persist.example.com`). The CA/Browser Forum Baseline Requirements define the same term as the bare FQDN used to obtain authorization (e.g., `example.com`). This conflict causes implementer confusion — especially once issue #33 adds an `adn` (Authorization Domain Name) field that uses the BR definition.

## Change

Rename the draft's concept to **"Validation Domain Name"** — aligning with RFC 8555 Section 8.4, which uses "validation domain name" for the analogous `_acme-challenge.<FQDN>` label in the dns-01 challenge. All 10 occurrences across 9 locations are updated:

| Section | Context |
|---|---|
| §2 Conventions | Glossary heading + definition body |
| §3 Challenge | Overview paragraph |
| §4 Challenge Response | Provisioning instructions (×2), example, server verification |
| §4.1 Multiple Issuer | CA verification step 1 |
| §4.3 JIT Validation | DNS lookup target |
| §6.8 Revocation | Record removal instructions |
| §8.2 Wildcard Example | Record provisioning step |

The glossary definition cites both the RFC 8555 precedent and the BR disambiguation.

## What this does NOT change

- No normative behavior changes — pure terminology rename
- No changes to record format, validation procedure, or security properties

## Prerequisite for

- **Issue #33** (subdomain validation prompting) — the `adn` field uses "Authorization Domain Name" in the BR sense; this rename clears the way

### DIFF
--- a/draft-ietf-acme-dns-persist.md
+++ b/draft-ietf-acme-dns-persist.md
@@ -89,8 +89,10 @@ Certification Authorities operating under various trust program requirements wil
 **DNS TXT Record Persistent DCV Domain Label**
 :   The label "_validation-persist" as specified in this document. This label is consistent with industry practices for persistent domain validation.
 
-**Authorization Domain Name**
+**Validation Domain Name**
 :   The domain name at which the validation TXT record is provisioned. It is formed by prepending the DNS TXT Record Persistent DCV Domain Label to the FQDN being validated.
+
+    This term follows the precedent set by {{!RFC8555}}, Section 8.4, which uses "validation domain name" for the analogous `_acme-challenge.<FQDN>` label in the dns-01 challenge. This document avoids the term "Authorization Domain Name" because the CA/Browser Forum Baseline Requirements {{cabf-br}} define it as the FQDN used to obtain authorization, without any label prepended.
 
 **Issuer Domain Name**
 :   A domain name disclosed by the CA in Section 4.2 of the CA's Certificate Policy and/or Certification Practices Statement to identify the CA for the purposes of this validation method.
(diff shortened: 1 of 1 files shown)


aarongable:
RFC 8555 doesn't define a term for this concept, but it does call it a "validation domain name" in one sentence (emphasis added):

> The record provisioned to the DNS contains the base64url encoding of
> this digest.  The client constructs the **validation domain name** by
> prepending the label "_acme-challenge" to the domain name being
> validated, then provisions a TXT record with the digest value under
> that name.

I slightly prefer "validation domain name" over "validation record name", but am happy either way.

Issue #56, opened 2026-04-02
Naming conventions


glatzert:
Hi everyone,

I'm maintainer of an ACME Server Implementation and I stumbled over the naming convention in this draft.

Take a look at (for example) https://www.rfc-editor.org/rfc/rfc8555#section-7.1.2 - naming is all camelCase and that's valid throughout RFC8555.
This draft adds new challenge objects and seems to follow no convention:

- `accounturi` neither matches the kebab-cased `issuer-domain-names` nor does it align with ACMEs camelCase (should probably be `accountUri`)
- `issuer-domain-names` does not align with ACMEs camelCase (should probably be `issuerDomainNames`)

sheurich:
@glatzert Good catch. The draft has two serialization contexts with different established conventions, and it follows neither cleanly.

DNS TXT record parameters use flat lowercase per RFC 8657/8659 convention: `accounturi`, `policy`. But `persistUntil` breaks the pattern: it should be `persistuntil` to match.

Challenge object JSON fields follow RFC 8555's camelCase: `validationRecord`, `caaIdentities`, etc. The current `issuer-domain-names` (kebab-case) and `accounturi` (flat lowercase) should be `issuerDomainNames` and `accountUri` to align.

That produces a deliberate split for the account identifier: `accountUri` in the JSON challenge object, `accounturi` in the DNS TXT record. Each format follows its own convention. The JSON field tells the client what value to place in the record's `accounturi` parameter; they are different fields in different serialization contexts.

I'll fix both in the next revision:

- DNS record parameters: `accounturi`, `policy`, `persistuntil` (flat lowercase, RFC 8657/8659 convention)
- Challenge object JSON: `accountUri`, `issuerDomainNames`, `persistUntil` (camelCase, RFC 8555 convention)

birgelee:
I don't think we can trivially fix "persistuntil" in the DNS record param. Its baked into the Baseline Requirements as "persistUntil" and right now we are drafting the I-D to be permitted for TLS server cert issuance via 3.2.2.4.22 DNS TXT Record with Persistent Value. If we move to "persistuntil" it breaks that and we need a CA/B F vote to fix it. Until we get alignment from CA/B F I suggest we keep the current language.

@slghtr-says do you have any insight on what the CA/B F's position is?

[...]

Issue #42, opened 2026-03-05
Don't limit validation reuse by TTL


aarongable:
The draft currently says

> The period for which a CA may rely on validation data is its Validation Data Reuse Period (as defined in {{conventions-and-definitions}}). However, if the DNS TXT record's Time-to-Live (TTL) is shorter than this period, the CA MUST treat the record's TTL as the effective validation data reuse period for that specific validation.

Given the other reuse period limitations in play (limits set by the BRs; limits set by persistUntil), it's confusing and potentially problematic to use the DNS TTL as a separate reuse limitation. As one example, it's common for authoritative nameservers to set very low TTLs by default. If that TTL is too low, it could be lower than the amount of time that it takes a CA to complete a validation and issuance cycle, effectively blocking issuance until the user changes the settings in their authoritative DNS.

It would be better to let DNS be DNS: the TTL can be whatever the user wants it to be, and that affects caching at the DNS layer. It doesn't affect validation reuse, which happens at a different layer (the CA's business logic). There's a potentially tricky interaction if a CA is using a caching recursive resolver, but that's fundamental to all validation techniques using DNS: the CA must configure any cache in their recursive resolver to be short enough that they can be confident their validations were performed within the acceptable window.

Also, relatedly, the client of a recursive resolver never actually sees the TTL of the record (as configured on the authoritative nameserver). The DNS client sees the TTL of the cached data in the recursive resolver if there is any. This is potentially quite confusing, since the validation reuse period will depend on a value that is not directly controlled by either the CA or the subscriber.

It would be better to omit this TTL-based limitation and instead let the relevant PKI rules (root programs, BRs) govern the validation reuse period.

sheurich:
@aarongable @birgelee Agreed. TTL and validation data reuse are independent layers that shouldn't be coupled. The practical problems reinforce that:

- Low default TTLs. Managed DNS providers commonly ship 60-300s defaults. A subscriber who hasn't tuned their TTL finds that dns-persist-01 validation can't be reused, which is the opposite of what this challenge type exists to do. A TTL of zero (common on dynamic records) triggers the current MUST NOT reuse language and reduces dns-persist-01 to plain dns-01 semantics.

[...]

Issue #39, opened 2026-02-26
Consider dedicated ACME error types for dns-persist-01 failure modes


sheurich:
## Problem

§9.3.1 maps three distinct failure modes to the same `unauthorized` error type:

| Failure | Remediation |
|---------|-------------|
| `accounturi` does not identify the requesting account | Reprovision with correct URI |
| `persistUntil` timestamp has expired | Reprovision with updated timestamp |
| `issuer-domain-name` does not match challenge | Reprovision with correct issuer |

Clients receiving `unauthorized` cannot programmatically distinguish these cases. The `detail` field may contain a human-readable description, but this is not machine-parseable and not all CAs will populate it consistently.

## Design question

Should this draft register its own error types in the ACME Error Types registry (RFC 8555 §9.6)?

For example:
- `urn:ietf:params:acme:error:persistExpired` — the `persistUntil` timestamp has passed
- Potentially: `accountMismatch`, `issuerMismatch`

### Arguments for

- Expiration is the failure mode most likely in steady-state operations (records that were valid last month silently become invalid). The remediation is clear and automatable.
- RFC 8555 registers purpose-specific error types (`caa`, `dns`, `rateLimited`, etc.) for the same reason: different failures need different responses.
- Pairs well with #38 — if the client didn't check locally, the CA error still tells it exactly what happened.

### Arguments against

- No ACME extension has registered its own error types yet. This draft would establish the precedent.
- Registering only `persistExpired` invites the question of why not the other two. Registering all three expands scope.
- If #33 (subdomain validation prompting) adds an `adn` field, new error cases (invalid ADN, ADN not a suffix) would also need types. Better to design the error type space once.
- §9.3.1 already encourages descriptive `detail` text. The incremental value of machine-readable types over descriptive strings is modest.

## Recommendation

Discuss scope and timing as a WG. This is a design question, not a bug fix. Filing for visibility and to capture the analysis before it's lost.

sheurich:
Deferring past `-02`. This is a WG design question, not a correctness bug: should dns-persist register validation-method-specific errors, and if so, should the registry cover only persistExpired or every validation failure? `-02` keeps the existing `malformed` and `unauthorized` guidance while the scope is settled.

PR #58, opened 2026-04-16
feat: support subscriber-account URIs in `accounturi`

### LINKED ISSUE #54: Pre-provisioning wording
> Domain owners MAY provision _validation-persist TXT records before any ACME interaction occurs.

Shouldn't this be more like

> Domain owners MAY provision _validation-persist TXT records before any ACME orders are created.

Domain owners would need to at least have created an ACME account in order to have an `accounturi` value for the pre-provisioned record.

### LINKED ISSUE #55: Confused about client validation of accounturi
With regards to Account URI Privacy, the alternative account URI is what would be presented by the CA in the `accounturi` field in the challenge object described in 3.1, right? But there's guidance in that description that says:

> Clients that pre-provisioned a record using their ACME account URL ([[RFC8555](https://www.ietf.org/archive/id/draft-ietf-acme-dns-persist-01.html#RFC8555)], Section 7.3) SHOULD verify the value identifies the same account.

But if it's an alternative URI, it purposefully won't match the real account URI and the client just has to trust that it's right, wouldn't it? Or am I missing where the alternative account URI comes from?

### LINKED ISSUE #57: Support subscriber-account URIs in `accounturi`
## Problem

The `accounturi` model currently requires each URI to "uniquely and permanently identify a single ACME account" ([§4.1 item 3](https://github.com/ietf-wg-acme/draft-ietf-acme-dns-persist/blob/df1b3e7e4cd8509294d20c38005d2a5e93422b96/draft-ietf-acme-dns-persist.md?plain=1#L167)). This works for per-account URLs and the 1:1 privacy URIs added in PR #35, but breaks down when a CA's subscribers manage large numbers of ACME accounts under a single relationship.

[Martijn Katerbarg](https://github.com/XolphinMartijn) (Sectigo) raised this on the [SC-098 discussion thread](https://groups.google.com/a/groups.cabforum.org/g/servercert-wg/c/Jyp40sfgQY8/m/NMhaRCyrCAAJ). SC-098 mandates CA processing of [RFC 8657](https://www.rfc-editor.org/rfc/rfc8657) CAA `accounturi` and `validationmethods` parameters by 2027-03-15. During that discussion, Martijn reported that individual Sectigo subscribers can have over 100,000 ACME accounts (per-server credentials). A subscriber wanting to use `accounturi` in CAA or dns-persist records would need a separate record per ACME account. That doesn't scale. SC-098 is being revised to accommodate this use case ([cabforum/servercert#567](https://github.com/cabforum/servercert/pull/567)); the account referencing scheme here is intended to be consistent with that ballot.

## RFC 8657 basis

RFC 8657 §3 already accommodates this for CAA. It defines "CA account" as an object "that represents a specific entity or **group of related entities**." The dns-persist draft intentionally narrowed this to "a single ACME account" for stronger security binding. The operational reality warrants relaxing that constraint, provided we add explicit safeguards and document the security tradeoffs.

## Design

Split §4.1 item 3 into three sub-items with distinct semantics:

- **(3a) ACME account URL** — MUST accept. Simple String Comparison. Unchanged.
- **(3b) Alternative 1:1 URIs** — MAY accept. Simple String Comparison or CA resolution. Privacy model from PR #35. Unchanged.
- **(3c) Subscriber-account URIs** — MAY accept. One URI maps to many ACME accounts, with five constraints:
  1. CA MUST verify the requesting ACME account is associated with the subscriber account through its own account management system.
  2. CA MUST document the URI format and association model in its CP/CPS.
  3. Same non-reassignment and deactivation lifecycle requirements apply.
  4. CA MUST log which specific ACME account was used for each validation against a subscriber URI.
  5. CA SHOULD make subscriber-account URIs syntactically distinguishable from per-account URIs.

No client-side protocol changes. Subscriber URIs are valid URIs provisioned identically to existing `accounturi` values. No new JSON fields, DNS record syntax, or IANA registrations.

## Security interactions

The 1:many model introduces nine interactions that need explicit treatment in the draft:

1. **`policy=wildcard` × subscriber URIs.** The broadest authorization the protocol can express: any ACME account under the subscriber, for any subdomain, for the record's lifetime. One compromised key out of 100K grants unlimited subdomain issuance.
2. **JIT validation × subscriber URIs.** JIT eliminates the challenge-response detection window. With subscriber URIs, any account under the subscriber gets JIT access, including accounts that didn't exist when the record was provisioned.
3. **Pre-provisioning × subscriber URIs.** ACME accounts created after record provisioning automatically inherit validation rights. The domain owner has no visibility into which accounts get added.
4. **Revocation model gap.** Deactivating one compromised ACME account doesn't invalidate the subscriber-level record for remaining accounts. The current emergency measure ("deactivate the ACME account") is insufficient.
5. **Comparison semantics.** Simple String Comparison doesn't apply to subscriber URIs. The CA performs an internal lookup instead.
6. **Privacy inversion.** Subscriber URIs increase correlation (one visible URI across all domains), the opposite of the 1:1 privacy model.
7. **DNS spoofing amplification.** The value of a spoofed dns-persist record to an attacker who also holds any ACME key under the subscriber is higher than with 1:1 binding. Amplifies the consequence of operating without DNSSEC.
8. **`adn` (issue #33) × subscriber URIs.** Once `adn` lands, it composes with `policy=wildcard` and subscriber URIs to deepen the scope of interaction #1. Forward reference needed.
9. **Subscriber credential compromise.** Compromise of the CA's subscriber account management system (portal, API) lets an attacker add new ACME accounts that immediately inherit all persistent authorizations. Distinct from ACME key compromise.

## Related issues

- Closes #55 — the 3a/3b/3c restructure resolves the confusion about client validation of alternative URIs by making the trust model explicit for each type
- Closes #54 — the pre-provisioning revision supersedes the narrow wording fix
- Related to #33 — interaction #8 cross-references `adn`
- Related to #39 — introduces a new failure mode but doesn't address error type registration

sheurich:
Fixes #57. Closes #55. Closes #54.

## Summary

Allow `accounturi` in dns-persist-01 records to identify a CA subscriber account encompassing multiple ACME accounts. See #57 for the problem statement, RFC 8657 basis, design rationale, and security analysis.

### §2: Subscriber Account definition

Defines "Subscriber Account" as a "CA account" (RFC 8657 §3) that encompasses one or more ACME accounts. Maps the term to the normative reference and explains the naming choice (disambiguation from ACME accounts, which are also CA accounts).

### §4.1 item 3: three-mode structure

| Mode                        | Cardinality | Comparison                                | Requirement |
|-----------------------------|-------------|-------------------------------------------|-------------|
| (3a) ACME account URL       | 1:1         | Simple String Comparison                  | MUST accept |
| (3b) Alternative URI        | many:1      | Simple String Comparison or CA resolution | MAY accept  |
| (3c) Subscriber-account URI | 1:many      | CA-internal lookup                        | MAY accept  |

Constraints on 3c:
1. CA MUST verify the requesting ACME account is associated with the subscriber account
2. CA MUST document the URI format and association model in CP/CPS
3. Non-reassignment and deactivation lifecycle requirements carry over
4. CA MUST log which specific ACME account was used for each subscriber-URI validation
5. CA SHOULD make subscriber URIs syntactically distinguishable from per-account URIs

Resolves #55: the trust model is now unambiguous for each URI type. Clients can independently verify 3a (string comparison), may be able to verify 3b (if the CA exposes the mapping), and cannot verify 3c (must trust the CA).

### §6.8: revocation rewrite

| Action                        | 3a (ACME URL)          | 3b (Alt URI)           | 3c (Subscriber)             |
|-------------------------------|------------------------|------------------------|-----------------------------|
| Remove DNS record             | ✓ (after cache expiry) | ✓ (after cache expiry) | ✓ (after cache expiry)      |
| Deactivate ACME account       | ✓ (immediate)          | ✓ (immediate)          | ✗ (only that account)       |
| Deactivate subscriber account | N/A                    | N/A                    | ✓ (immediate, all accounts) |

CAs supporting subscriber-account URIs MUST provide a subscriber-account deactivation mechanism.

### §7.2: account binding security

- Existing claim that security is "fundamentally bound to the security of the ACME account's private key" is incomplete for subscriber URIs — bound to any key under the subscriber plus the subscriber account management system.
- New subsection (§7.2.1) on subscriber credential compromise: compromise of the CA's account management system lets an attacker add ACME accounts that immediately inherit all persistent authorizations.

### Security considerations

Addresses all nine interactions from #57:
- Interactions 1–3 and 6–8: new subsections or additions in Security Considerations
- Interaction 4: addressed by §6.8 rewrite
- Interaction 5: addressed by §4.1 restructure
- Interaction 9: addressed by new §7.2.1 subsection

### Other text updates

- §1: qualify "uniquely identifies the applicant's account" to cover subscriber accounts
- §3.1: broaden `accounturi` field description; add per-mode client verification guidance
- §4.4: subscriber URI as pre-provisioning source; forward-authorization advisory
- Case-handling rules: Simple String Comparison for 3a/3b; CA-internal verification for 3c
- Error handling (§9.3.1): cover the subscriber-association failure mode
- §7.6 CAA interaction: use `dns-persist-01` consistently; note CAA and dns-persist `accounturi` as independent reinforcing checks
- Quote RFC 8657 §3 "group of related entities" directly in normative text

## Deferred

- Client mismatch behavior (#59) — warrants focused treatment per mode
- IANA error type registration (#39) — new failure mode noted but not registered
- `adn` response field (#33) — forward-referenced only
- Naming conventions (#56) — independent editorial issue
- TTL validation reuse (#42) — independent issue

No new JSON fields, DNS record syntax, or IANA registrations. The change is purely semantic.


draft-ietf-acme-dns-persist.md
```diff
@@ -120,7 +120,7 @@ The challenge object for "dns-persist-01" contains the following fields:
...
+- **accounturi** (required, string): A URI identifying the specific ACME account requesting validation or a subscriber account encompassing it (see {{validation-record-format}}), using the identifier format specified in {{!RFC8657}}, Section 3. This is the URI the CA expects in the DNS record and the mechanism by which the CA communicates alternative URIs to the client. Clients that pre-provisioned a record using their ACME account URL ({{!RFC8555}}, Section 7.3) SHOULD verify the value identifies the same account. Note that independent verification is possible for mode 3a (Simple String Comparison), may be possible for mode 3b (if the CA exposes the URI-to-account mapping), and is not possible for mode 3c (clients must trust the CA's subscriber-account association).
```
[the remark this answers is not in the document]
sheurich: Valid gap. The SHOULD verify clause predates this PR but adding the per-mode verification note makes the gap more visible. Mismatch behavior interacts differently across modes: for 3a, string mismatch is definitive; for 3b, it depends on whether the client can resolve the mapping; for 3c, the client cannot verify at all. That complexity warrants focused treatment rather than a sentence grafted onto this change.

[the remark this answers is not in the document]
sheurich: Tracked in https://github.com/ietf-wg-acme/draft-ietf-acme-dns-persist/issues/59

draft-ietf-acme-dns-persist.md
```diff
@@ -156,9 +159,23 @@ The RDATA of this TXT record MUST fulfill the following requirements:
...
 
-3.  The issue-value MUST contain an `accounturi` parameter whose value is a URI that identifies the ACME account requesting validation.
+3.  The issue-value MUST contain an `accounturi` parameter whose value is a URI that identifies the ACME account requesting validation, either directly or through a subscriber-account association.
+
+    The `accounturi` value MUST conform to one of the following three modes:
+
+    *   **(3a) ACME account URL**: CAs MUST accept the URI of the ACME account object ({{!RFC8555}}, Section 7.3) as a valid `accounturi` value.
+
+    *   **(3b) Alternative 1:1 URI**: CAs MAY accept other URIs, provided each such URI uniquely and permanently identifies a single ACME account and satisfies the uniqueness requirements of {{!RFC8657}}, Section 5.4. CAs MUST NOT reassign a URI to a different account. When an ACME account is deactivated, CAs MUST treat all 1:1 URIs associated with that account as invalid for validation purposes.
```
birgelee: I think we are on the same page with what this method is in tended imply, but I wonder if using "1:1" will confuse people. I would expect a 1:1 to mean that there is exactly 1 alternative URI mapping back to an ACME account and there can exist no other alternative mapping strategies.

I understand the language is not intended to mean this (the next sentence specifically says "URIs" implying there can be multiple mappings), but I had always thought of this as a many-to-one not 1:1 style mapping.

We could also just drop the "1:1" reference together and just call this something like "Alternative URI**"

birgelee (commented):
I like it, would just remove the "1:1" reference to avoid confusion.

sheurich:
Re [Henry's comment on "1:1"](https://github.com/ietf-wg-acme/draft-ietf-acme-dns-persist/pull/58#discussion_r2106786654): dropped "1:1" from the mode 3b label and all six references. The normative text ("each such URI uniquely and permanently identifies a single ACME account") already captures the constraint unambiguously.

birgelee:
@sheurich thanks for updating that. This is good to go from my perspective.

draft-ietf-acme-dns-persist.md
```diff
@@ -156,9 +159,23 @@ The RDATA of this TXT record MUST fulfill the following requirements:
 
 2.  The `issuer-domain-name` portion of the issue-value MUST be one of the Issuer Domain Names provided by the CA in the `issuer-domain-names` array of the challenge object. If the `issuer-domain-name` does not match any of the provided values, the CA MUST reject the record.
 
-3.  The issue-value MUST contain an `accounturi` parameter whose value is a URI that identifies the ACME account requesting validation.
+3.  The issue-value MUST contain an `accounturi` parameter whose value is a URI that identifies the ACME account requesting validation, either directly or through a subscriber-account association.
+
+    The `accounturi` value MUST conform to one of the following three modes:
+
+    *   **(3a) ACME account URL**: CAs MUST accept the URI of the ACME account object ({{!RFC8555}}, Section 7.3) as a valid `accounturi` value.
```
MaxHearnden: This would imply that CAs that support HTTP reverse proxies (which is currently allowed) would accept arbitrary URI origins.

draft-ietf-acme-dns-persist.md
```diff
@@ -391,12 +426,20 @@ DNS records are generally not authenticated end-to-end, making them potentially
 
 Additionally, CAs SHOULD protect their `issuer-domain-name` with appropriate security measures. Using DNSSEC to protect the CA's `issuer-domain-name` is RECOMMENDED. An attacker who compromises the DNS for a CA's `issuer-domain-name` could disrupt validation or potentially impersonate the CA in certain scenarios. While this is a systemic DNS security risk that extends beyond this specification, it is amplified by any mechanism that relies on DNS for identity.
 
+### DNS Spoofing Amplification with Subscriber URIs {#dns-spoofing-amplification-with-subscriber-uris}
+
+The value of a spoofed `_validation-persist` TXT record to an attacker is significantly amplified when the spoofed record uses a Subscriber-account URI. If an attacker manages to spoof a DNS response containing a Subscriber-account URI, they can use *any* compromised ACME key associated with that subscriber to issue certificates for the spoofed domain. This amplification of risk further emphasizes the necessity of strong DNS security measures, such as DNSSEC, when operating ACME infrastructure.
```
MaxHearnden, line 479: Why does the ACME key need to be compromised for this attack? Surely the attacker could legitimately create their own ACME account.

draft-ietf-acme-dns-persist.md
rmbolger: I don't follow. The account URI in the record doesn't have to actually be usable for real (like have a web server able to respond to it). It's just an arbitrary string that should map to an account from the CA's perspective.

MaxHearnden: The issue is that a CA operating under mode 3a would be prohibited from requiring that the URI uniquely identifies the account.

draft-ietf-acme-dns-persist.md
rmbolger: Perhaps this is more relevant for commercial providers where it's not as easy to create an attacker controlled account? But even if the attacker can spoof DNS responses with a victim's Subscriber-account URI, the attacker still needs to have compromised one of the victim's account keys in order to sign the ACME requests with, right?

draft-ietf-acme-dns-persist.md
rmbolger: Wait what? 3a is already how ACME CAs operate today where accounts have a single URI value that uniquely identifies the account. Why would they be prohibited from continuing to do that?

MaxHearnden: The issue is that an RFC8555 account URI doesn't uniquely identify the account and CAs like Let's Encrypt allow the domain name in an account URI to be changed while still referring to the same account. This can cause a security issue with this challenge type as a valid account URI might point to an attacker controlled domain.

rmbolger: Sure it does. An account URI today only refers to a single account on the CA side. If the CA changes (for example) the domain in the account URI, the new URI still only refers to a single account on the CA side. You just now have two account URIs that uniquely identify an account (assuming the CA still recognizes the old one).

Perhaps we're interpreting the phrase "uniquely identify the account" differently? I take it to mean that at any given time, an account URI can refer to only one account on the CA side. There can theoretically be many account URI values that refer to the same account. But they can all only refer to a single account.

birgelee: on this point I will mention that this is the existing behavior inherited from RFC 8657 which shares many of these same security considerations. Additionally, if we assume the CA and the subscriber are benign, it seems unlikely the subscriber will intentionally put in an arbitrary adversary-controlled prefix to their accounturi. A goal for this standard is to maintain consistency with RFC 8657, so I am somewhat keep this language as is.

draft-ietf-acme-dns-persist.md
birgelee: The context for this is imagine a CA which has some type of non-ACME subscriber relationship (e.g., an enterprise account) which contains multiple ACME accounts within it (say an ACME account for each ACME client within the enterprise). These associated ACME accounts are understood by the CA and only the enterprise administrator can make new ACME accounts within the enterprise.

CAB/F participants have asked to let this method support specifying the account URI of the **enterprise subscriber** account and permit this to authorize issuance by the sub ACME accounts. One security consideration of this approach is that it creates a larger attack surface because there are now N different authorized ACME accounts and a compromise of the private key of any one of those accounts will authorize issuance. IMO the alternative of forcing a single ACME account for the entire enterprise will just mean the same ACME private key will be spread over N different servers, so I think this is a good approach to facilitate enterprise use cases.

My understanding of this text is that its introduced to help explain this security consideration. I will say I find the combination of DNS record spoofing and this enterprise account attack surface expansion somewhat confusing as if the attacker is capable of spoofing a DNS record, the obvious attack is to just spoof their own account URI in the record. My proposal would be to remove the words spoofing and focus the section on the expanded attack surface that comes from enterprise level subscriber accounts.

draft-ietf-acme-dns-persist.md
MaxHearnden: RFC 8657 only has a MAY requirement on accepting ACME account URIs (https://www.rfc-editor.org/rfc/rfc8657#section-3.1-1) instead of the MUST requirement here.

I think what I've been trying to say is that given dns-persist-01's reliance on HTTPS, CAs should be given the option to reject some account URIs that they otherwise allow for other methods e.g. Let's Encrypt should be allowed to reject http://invalid.example/acme/acct/3285011975 as an account URI for the purposes of this method.

sheurich: @MaxHearnden agreed, the MUST is stronger than warranted. We're revising mode 3a to SHOULD with a soft-redirect:

> (3a) ACME account URL: CAs SHOULD accept the URI of the ACME account object (RFC 8555, §7.3) as a valid `accounturi` value. A CA that does not accept the account URL directly MUST provide an alternative URI in the challenge `accounturi` field (mode 3b or 3c).

Reasons why this is the right level:

1. RFC 8657 §5.2 gives CAs explicit discretion over which parameters to recognize and how to interpret them. §5.3 goes further in that a CA that can't support the parameters uniformly MAY split into multiple CAA-processing identities, and one that can't even do that MUST NOT implement the parameters. A MUST on mode 3a overrides that design.

2. The challenge-response `accounturi` field (§3.1) already works as a substitution mechanism, as the CA can return a different URI than what the client proposed. A MUST on accepting the raw account URL makes that channel pointless for mode 3a. SHOULD preserves it.

The soft-redirect ensures subscribers always get a working `accounturi` regardless of which modes a given CA supports.

@birgelee the SHOULD preserves §5.2's variance model while the redirect keeps subscribers from hitting a dead end.

draft-ietf-acme-dns-persist.md
sheurich: Agreed, the DNS-spoofing framing conflates two threats and makes the section harder to follow. The concern specific to §7.2.1 is attack surface expansion: N authorized ACME accounts under one subscriber relationship means N private keys that could lead to unauthorized issuance. DNS manipulation is orthogonal and already covered under MPIC.

The attacker doesn't need to compromise an existing key — they need access to the subscriber's account management system to add a new account. The current §7.2.1 text says this, but the parent section muddies it by leading with key compromise scenarios.

I'll revise §7.2 and §7.2.1:

1. Separate key compromise (§7.2 parent, all modes) from subscriber-management compromise (§7.2.1, mode 3c).
2. Drop DNS spoofing references — orthogonal, already addressed under MPIC.
3. Focus §7.2.1 on the enterprise attack surface expansion per @birgelee's proposal.

PR #62, opened 2026-05-07
Add procedure for accounturi verification

MaxHearnden:
This PR adds a procedure for ACME clients to verify accounturis. The important step is to make a POST request to the accounturi endpoint, the token field/parameter is just so that ACME servers can force clients to implement the verification process.

This also requires the accounturi to be a valid ACME endpoint however I wasn't sure where it should be stated in the draft.

### DIFF
--- a/draft-ietf-acme-dns-persist.md
+++ b/draft-ietf-acme-dns-persist.md
@@ -148,6 +148,14 @@ For example, if the domain being validated is "example.com", the Validation Doma
 
 The client indicates it is ready for validation by POSTing an empty JSON object (`{}`) to the challenge URL, following the procedure defined in {{!RFC8555}}, Section 7.5.1.
 
+## Account URI verification {#account-uri-verification}
+
+The client verifies the provided accounturi by POSTing an empty JSON object (`{}`) to the accounturi URL, following the procedure defined in {{!RFC8555}}, Section 7.5.1.
+
+If the request was not successful then the challenge MUST be aborted.
+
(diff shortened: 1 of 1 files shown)


sebastiannielsen:
My suggestion is that we should not overload the server with more requests that must be done.

A better solution would be as I have suggested, allow the client to compute the "accounturi" value offline by using a pubkey: "protocol" which is the sha256 value of the public key computed using the SPKI format over the account key - which also makes it crypto agility since the algoritm is included in the SPKI.

Meaning, a "TLSA" record calculator over the account public key would give the correct sha256 value.

Like:
_validation-persist.yourdomain.tld IN TXT "letsencrypt.org;accounturi=pubkey:fa759db3816ab7387006c8fb5b887a6fc3dde2d77f593ec2a8bf6504524da26e"

or use the JWK thumbprint.

By permitting a offline-calculation, all security considerations with a compromised accounturi is erased - at the cost of requiring record update if you choose to rotate account keys.

sebastiannielsen:
I propose to reject this pull with the explanation that adding requirements that increases the number of a requests a subscriber has to do to the ACME server to facilitate a specific validation method should be avoided as it increases load on CA infrastructure.

Especially when it can be solved by this pull instead: https://github.com/ietf-wg-acme/draft-ietf-acme-dns-persist/pull/65

at the cost of having to replace the record when doing a account key rollover.

birgelee:
Yeah I am somewhat in favor of just locking down the syntax (see my comment on 65) to avoid needing the server and client to talk more.

MaxHearnden:
After considering the use cases for this procedure, I think that it might be worth pursuing this for mode 3c account URIs (multi-account), unless 3c account URIs will only be provided out of band.

Issue #61, opened 2026-04-30
accounturi privacy: mechanism design


sheurich:
# The accounturi privacy mechanism

This issue hosts the design discussion the WG list deferred per the [2026-04-24 consensus call][consensus-call].

[consensus-call]: https://mailarchive.ietf.org/arch/msg/acme/AMZxwvkdtXPIttRhFqQvL-2sAco/

## Context

On [2026-04-24][consensus-call] a consensus call opened on whether static ACME account URIs in `_validation-persist` TXT records are a privacy concern this draft should address. The records leak which domains share an ACME account to anyone who queries DNS. The [2026-04-23 analysis][ounsworth-apr23] laid out the leak mechanics.

The list thread surfaced one constraint, that the `accounturi` space is not high-entropy, per [Seo Suchan, 2026-04-24][seo-apr24], and one candidate construction, `SHA256(accounturi || domain_name)`, that fails under it.

Since the consensus call, several participants' positions have converged. Henry Birge-Lee [posted a detailed analysis][birgelee-apr27] showing that SHA-256 over the LE account-URI space breaks in ~10 seconds on a consumer GPU, independently confirming Seo's constraint. He proposed a CA-side lookup table as the most practical approach. Sebastian Robin Nielsen [pushed back][nielsen-apr27-2] with a `sha256(account_public_key || domain)` counter-proposal; Henry [distinguished][birgelee-apr27-reply] the two CDN threat scenarios (shared-owner/separate-infrastructure vs. shared-owner/shared-infrastructure) and defended the privacy case. Mike Ounsworth subsequently [endorsed flexibility over mandate][ounsworth-apr29]. Recent list discussion suggests alignment on three points: the privacy concern is real for a class of subscribers, MAY flexibility is appropriate, and public-input constructions are broken.

Scope: unlinkability mechanism for `accounturi` in `_validation-persist` records, and the normative posture (MUST / SHOULD / MAY). Out of scope: substitution attacks on `accounturi` (see #58); subscriber-URI semantics (proposed in #58); verification binding via account-key thumbprints (see "Rejected classes").

[ounsworth-apr23]: https://mailarchive.ietf.org/arch/msg/acme/g9iIiRMDL2X31cgM55-cg3SzV3M/
[seo-apr24]: https://mailarchive.ietf.org/arch/msg/acme/Ylp0_e2bFxX_D-YyxK6WWki-pY8/

## Terminology

Mode labels used in this issue:

[...]

PR #65, opened 2026-05-17
new mode to allow for public key.

### LINKED ISSUE #64: Record should include client-computed information
Having the dns-persist-01 record contain only information provided by the ACME server opens the possibility for Meddler-in-the-Middle attacks. The attack flow would be something like:
- Set up a malicious ACME server
- Get a victim to configure your server's directory URL in their ACME client
- When the victim registers an account with your server, register a new account with a real ACME server, and return _that_ account's URI verbatim
- When the victim provisions a dns-persist-01 record, the record will actually authorize the attacker's ACME account at the real CA; the attacker can now issue arbitrary certificates for the victim's domain

The original RFC 8555 discusses this exact attack in its [Security Considerations section](https://datatracker.ietf.org/doc/html/rfc8555#section-10.2). The mitigation used by those original ACME challenges is that the token the Applicant needs to present consists of two parts:
1. The random value provided by the server; and
2. A hash of the applicant's ACME account public key

Importantly, (2) is computed entirely offline -- both the ACME client and the ACME server compute it independently. And while the malicious actor _can_ compute the hash of the victim's pubkey (it's public, after all), the malicious actor _can't_ convince the ACME server that that's the Key Authorization it should be looking for. When the malicious actor tries to get a cert for the victim's domain from the real ACME server, that server will expect to see a hash of the _attacker's_ pubkey, and there's no reason for the victim to have provisioned that value.

Thus, I believe that dns-persist-01, in order to mitigate this same attack, must include a value computed over the applicant's ACME client pubkey.

I will note that this introduces a tradeoff: improve security by keeping DNS API credentials far away from webservers and ACME clients, but impair security by preventing applicants from rotating their ACME client keys. If anyone has suggestions for how to prevent this attack without also preventing key rotation, I'd love to hear it.

See this thread on the ACME mailing list for further discussion: https://mailarchive.ietf.org/arch/msg/acme/i_kd1g7vFaWO8x3_Ukh6Iz3nems/

sebastiannielsen:
Closes https://github.com/ietf-wg-acme/draft-ietf-acme-dns-persist/issues/64 if merged.

### DIFF
--- a/draft-ietf-acme-dns-persist.md
+++ b/draft-ietf-acme-dns-persist.md
@@ -175,7 +175,7 @@ The RDATA of this TXT record MUST fulfill the following requirements:
         4.  The CA MUST log which specific ACME account was used for each validation against a subscriber URI.
         5.  The CA SHOULD make subscriber-account URIs syntactically distinguishable from per-account URIs. If subscriber-account URIs are not distinguishable, clients and auditors cannot independently determine which verification model applies to a given `accounturi` value.
    
-    *   **(3d) ACME Public key**: CAs MAY accept a URI that starts with pubkey: and contains a Base64_URL encoded sha256 hash of the subscriber JWK public key as described in ({{!RFC7638}}, Section 3). If a CA supports this mode, it MUST enforce the following constraints:
+    *   **(3d) ACME Public key**: CAs MAY accept a URI that starts with pubkey: and contains a Base64_URL encoded sha256 hash of the account JWK public key as described in ({{!RFC7638}}, Section 3). If a CA supports this mode, it MUST enforce the following constraints:
 
         1.  The CA MUST verify that the requesting ACME account's public key is currently associated with the account in question and the base64_URL hash value matches the value after the pubkey: identifier.
         2.  The CA MUST deactivate any authorizations that have been validated using this mode if a successful keyChange as descibed in ({{!RFC8555}}, Section 7.3.5) is made.
--- a/draft-ietf-acme-dns-persist.md
+++ b/draft-ietf-acme-dns-persist.md
@@ -178,7 +178,7 @@ The RDATA of this TXT record MUST fulfill the following requirements:
     *   **(3d) ACME Public key**: CAs MAY accept a URI that starts with pubkey: and contains a Base64_URL encoded sha256 hash of the account JWK public key as described in ({{!RFC7638}}, Section 3). If a CA supports this mode, it MUST enforce the following constraints:
 
         1.  The CA MUST verify that the requesting ACME account's public key is currently associated with the account in question and the base64_URL hash value matches the value after the pubkey: identifier.
-        2.  The CA MUST deactivate any authorizations that have been validated using this mode if a successful keyChange as descibed in ({{!RFC8555}}, Section 7.3.5) is made.
+        2.  The CA MUST deactivate any authorizations that have been validated using this mode if a successful keyChange as described in ({{!RFC8555}}, Section 7.3.5) is made.
         3.  The CA MUST permit, in this mode, both invalid and valid records to co-exist, to allow for a key rollover where the new record is published prior to when a new key is taken into production.
         4.  The CA MUST reassign a public key to a new ACME account, provided the old account performs the keyChange action as described in ({{!RFC8555}}, Section 7.3.5) to take the key OFF the old account, and the new account subsequently performs the keyChange action to install the key ON the new ACME account. If a CA does NOT provide the keyChange endpoint in its directory as described in ({{!RFC8555}}, Section 7.1.1), the CA MUST NOT reassign a public key to a new ACME account.
 
--- a/draft-ietf-acme-dns-persist.md
+++ b/draft-ietf-acme-dns-persist.md
@@ -413,7 +413,7 @@ creates a verifiability gradient:
 - **Mode 3d (ACME Public key):** The client can verify the provisioned
   public key offline by performing the neccessary calculations offline.
   This can be done even if the client for the moment lacks network
-  connectiviity, for example before DHCP has been completed or before
+  connectivity, for example before DHCP has been completed or before
   802.11x authentication to the local network. 
 
 This gradient means that a CA operating in mode 3b or 3c can direct a client
--- a/draft-ietf-acme-dns-persist.md
+++ b/draft-ietf-acme-dns-persist.md
@@ -174,6 +174,13 @@ The RDATA of this TXT record MUST fulfill the following requirements:
         3.  The CA MUST NOT reassign a subscriber-account URI to a different subscriber. When a subscriber account is deactivated, the CA MUST treat the subscriber-account URI as invalid for validation purposes.
         4.  The CA MUST log which specific ACME account was used for each validation against a subscriber URI.
         5.  The CA SHOULD make subscriber-account URIs syntactically distinguishable from per-account URIs. If subscriber-account URIs are not distinguishable, clients and auditors cannot independently determine which verification model applies to a given `accounturi` value.
+   
+    *   **(3d) ACME Public key**: CAs MAY accept a URI that starts with pubkey: and contains a Base64_URL encoded sha256 hash of the subscriber JWK public key as described in ({{!RFC7638}}, Section 3). If a CA supports this mode, it MUST enforce the following constraints:
+
+        1.  The CA MUST verify that the requesting ACME account's public key is currently associated with the account in question and the base64_URL hash value matches the value after the pubkey: identifier.
+        2.  The CA MUST deactivate any authorizations that have been validated using this mode if a successful keyChange as descibed in ({{!RFC8555}}, Section 7.3.5) is made.
+        3.  The CA MUST permit, in this mode, both invalid and valid records to co-exist, to allow for a key rollover where the new record is published prior to when a new key is taken into production.
+        4.  The CA MUST reassign a public key to a new ACME account, provided the old account performs the keyChange action as described in ({{!RFC8555}}, Section 7.3.5) to take the key OFF the old account, and the new account subsequently performs the keyChange action to install the key ON the new ACME account. If a CA does NOT provide the keyChange endpoint in its directory as described in ({{!RFC8555}}, Section 7.1.1), the CA MUST NOT reassign a public key to a new ACME account.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-dns-persist.md
```diff
@@ -174,6 +174,13 @@ The RDATA of this TXT record MUST fulfill the following requirements:
...
         4.  The CA MUST log which specific ACME account was used for each validation against a subscriber URI.
         5.  The CA SHOULD make subscriber-account URIs syntactically distinguishable from per-account URIs. If subscriber-account URIs are not distinguishable, clients and auditors cannot independently determine which verification model applies to a given `accounturi` value.
+   
+    *   **(3d) ACME Public key**: CAs MAY accept a URI that starts with pubkey: and contains a Base64_URL encoded sha256 hash of the account JWK public key as described in ({{!RFC7638}}, Section 3). If a CA supports this mode, it MUST enforce the following constraints:
+
+        1.  The CA MUST verify that the requesting ACME account's public key is currently associated with the account in question and the base64_URL hash value matches the value after the pubkey: identifier.
+        2.  The CA MUST deactivate any authorizations that have been validated using this mode if a successful keyChange as described in ({{!RFC8555}}, Section 7.3.5) is made.
```
MaxHearnden: This would violate section 7.3.5 of RFC 8555 (https://www.rfc-editor.org/rfc/rfc8555#section-7.3.5) which states that
```
   Note that changing the account key for an account SHOULD NOT have any
   other impact on the account.  For example, the server MUST NOT
   invalidate pending orders or authorization transactions based on a
   change of account key.
```

sebastiannielsen: note "pending". Meaning that completed authorizations may be invalidated for security purposes.

The security risk of leaving a valid authorization, after a keyChange has been performed, is that two accounts may be authorized to issue for a single DNS record, which violates CAB/F rules that explicitly state that each DNS record may only reference a single CA account.

So a pending authorization would still be valid, since the challenge has not been completed. A completed authorization would however be invalidated, which can result in a pending order being invalidated - but that does not mean the order was invalidated as a consequence of a keyChange, but a dependent (completed) authorization.

The reason to invalidate a authorization is also that the security status of the account has been changed, which may warrant the CA changing the status of completed authorizations or challenges, as they are no longer valid for issuance at issuance-time.

The rationale of not invalidating pending transactions is to not break the state in a client and thus cause a desync in the client, where a client believes the pending authorizations are still valid, and try to complete them, even tough the server have changed state. **For authorization reuse, the client re-checks the state of authorizations, and would see the state change and update its local state, so the risk of invalidting a completed authorization is less of a risk - because a CA may invalidate a authorization for other purposes (mississuance, etc) so the client MUST recheck a authorization before reusing its status.**

sebastiannielsen: Changed the pull to explicitly prohibit invalidating authorizations that have another state than "valid".

draft-ietf-acme-dns-persist.md
```diff
@@ -174,6 +174,13 @@ The RDATA of this TXT record MUST fulfill the following requirements:
...
+        2.  The CA MUST deactivate any authorizations that have been validated using this mode if a successful keyChange as described in ({{!RFC8555}}, Section 7.3.5) is made. Note that a CA MUST NOT invalidate an authorization that has another state than "valid" as a result of a keyChange, as this would violate ({{!RFC8555}}, Section 7.3.5).
+        3.  The CA MUST permit, in this mode, both invalid and valid records to co-exist, to allow for a key rollover where the new record is published prior to when a new key is taken into production.
+        4.  The CA MUST reassign a public key to a new ACME account, provided the old account performs the keyChange action as described in ({{!RFC8555}}, Section 7.3.5) to take the key OFF the old account, and the new account subsequently performs the keyChange action to install the key ON the new ACME account. If a CA does NOT provide the keyChange endpoint in its directory as described in ({{!RFC8555}}, Section 7.1.1), the CA MUST NOT reassign a public key to a new ACME account.
```
MaxHearnden, line 183: Why is this a MUST? CAs might decide that once a key has been used for an account, it can't be reused again for a different account.

sebastiannielsen: I feel that a key should be able to be transferred between accounts at will of the subscriber. a CA should only blacklist a key if it believes the key has been compromised - OR if the key is already assigned to another account (ergo in use - 2 accounts may never share the same key ever at the same time).

Thats why I suggested the point 2 to avoid a CA falling into violation of the CAB/F rules by disallowing a authorization based on the old key to remain valid for issuance after the key has been moved to a new account.

There can be several reasons a subscriber wants to move a key to another account - for example if a subscriber sells his company but want to remain in control of other domains provisioned on the same key, - or an user leaving a web hotel where the web hotel manages the ACME accounts.

This doesn't prevent a CA from invalidating account keys that may be sensitive, or invalidating account keys as a result of a company going bankrupt. (to prevent computers and servers having the keys from being abused after they have been sold in a bankrupty auction).

**What the MUST says is that the keyChange action may never cause a key to be blacklisted in itself. If a subscriber intentionally want to blacklist a key, they should invalidate the account instead, which will invalidate the key aswell forever.**

This creates 2 distinct actions - either you keyChange to keep the old key usable, or you invalidate account if you want to also block the old key.

Theres little value in keeping the account itself, especially now when the new CAB/F rules that state that authorizations may only be valid for 7 hours when certificate validity goes down to 45 days.

_Another reason to wanting to go back to a old key, is if a key rollover fails spectacularly, and the subscriber want to roll back the key and retry again at a later time.
Blacklisting a key because it has been "used" would cause the subscriber to not be able to revert the key rollover if something goes south._

Of course this means, that keys that may leak after a rollover could still be usable - thats why its important that a subscriber removes any old previous DNS-PERSIST-01 pubkey: records after having completed a key rollover process and also have verified everything works with the new key.

birgelee:
Hi all,

We are moving to decline this PR primarily on implementation grounds. We do want to move forward with a public-key hash mechanism, but we would like to take it in a slightly different direction. Internally we have aligned on the following format for hashing:

SHA-256(pubkey || accountURL || domain)

(|| for concatenation) This hash strategy has strong alignment with what is already required by the CA/Browser Forum and provides stronger security properties than pubkey alone.

Essentially, the justification for putting both the standard account URL and the pubkey in the hash is that the pub key provides sufficient entropy to avoid the account URL enumeration hashing attack (i.e., just iterating through account URLs until finding one that hashes to the observed value) while the account URL ensures CA/B Forum compliance and avoids complications with multiple accounts that share keys.

The current wording from the CA/B Forum reads "The `issue-value` MUST contain an `accounturi` parameter, where the parameter value is a unique URI (as described by RFC 8657, Section 3) identifying the account of the Applicant which requested validation for this FQDN; ". I feel that SHA-256(pubkey || accountURL || domain) satisfies this criteria because of the collision-resistant properties of hashes. There can only be one input that produces this hash so the hashed value can serve to uniquely identify the accountURL. SHA-256(pubkey) does not provide this same guarantee as two accounts could use the same key (potentially even between CAs). Obviously additional restrictions in the ecosystem like constraints on accounts reusing keys could be enforced, but I would argue that this comes at the cost of significantly more complexity (and could be a security risk, known bad keys could be tried during account registration to see if they are used by an ACME account).

In addition to stronger CA/B Forum language compliance, SHA-256(pubkey || accountURL || domain) solves the reused pubkey problem in a much more head on manner. Key reuse between accounts (or even accounts at different CAs) never risks overlaps in DCV.

Finally, we felt there are some advantages to keeping it in the URL format. Although the URI specifier can be much more broad, keeping it to a URL simplifies string comparison (its always byte-for-byte identical), and has nice multi-CA namespace isolation. Right now I am considering a scheme somewhat like:

https://<ca directory URL domain name>/.well-known/account-hash/<SHA-256(pubkey || accountURL || domain)>


Thank you again for this PR. I do think it is helping us move in the right direction by approaching these problems more head on, but we would prefer not to go with a special-purpose "pubkey" URI format.

MaxHearnden:
> Essentially, the justification for putting both the standard account URL and the pubkey in the hash is that the pub key provides sufficient entropy to avoid the account URL enumeration hashing attack (i.e., just iterating through account URLs until finding one that hashes to the observed value) while the account URL ensures CA/B Forum compliance and avoids complications with multiple accounts that share keys.

Just to restate my issue with the privacy claim.

As I had stated in https://github.com/ietf-wg-acme/draft-ietf-acme-dns-persist/issues/61#issuecomment-4443182076, using SHA-256 as proposed still allows an attacker to determine if two domains were issued from the same account. Furthermore given knowledge of an account URI (maybe from a single CAA record or from brute forcing two domains) and a single SHA-256(pubkey || accountURL || domain) one can quickly (with a single hash operation) determine if that account is used for another domain.

birgelee:
@MaxHearnden am I correct in understanding that your attack is based on the assumption that the adversary knows the account URI or at least has a candidate account URI in mind that they want to test? I know you point out account URIs can be leaked via CAA record's accounturi parameter, but if the account URI is being leaked through this channel, I feel the adversary could perform the same infrastructure linkage through the CAA accounturi parameter.

I see the goal of this system I think of as mostly to prevent bulk infrastructure linkage through internet-wide scanning done through platforms like Shodan and Censys. If an adversary digs deep I feel shared infrastructure can usually be uncovered (it was discussed on list if this entire privacy system had a point because of this reason) and if a victim's account URI is leaked I feel that is out of scope for this mechanism.

I think even pairwise attacks (i.e., a function that can return True/False on two specific to tell if they use the same account URI) are sort of less of a concern as well. I think the only real risk comes from plaintext accountURIs as an adversary can just cheaply search for them in other places.

With that said, there is a potential approach of using an HMAC where the CA shares an HMAC key (say k) with the client (during new order or ideally account provisioning) and the client computes HMAC_k(pubkey || accountURL || domain). Collision resistance of hashes means that even if an adversary is between the CA and the ACME client, the adversary cannot construct a k' such that

HMAC_k(pubkey_adversary || accountURL_adversary || domain) = HMAC_k'(pubkey_victim || accountURL_victim || domain)

Since the adversary could never convince the victim to upload the HMAC to its account, this would still stop the MitM attack.

That said, this would hurt pre-provisioning unless it was communicated at account registration time which to me seems like an expensive lift for this system.

PR #66, opened 2026-05-17
Change language since issuer-domain-name is a opaque value

### LINKED ISSUE #63: Suggestion: Change wording of the "Record tampering" to not link to issuer-domain-name
I suggest the following:

```
DNS records are generally not authenticated end-to-end, making them potentially vulnerable to tampering. CAs SHOULD implement additional integrity checks where possible and consider the overall security posture of the DNS infrastructure when relying on persistent validation records.

Additionally, CAs SHOULD protect their issuer-domain-name with appropriate security measures. Using DNSSEC to protect the CA's issuer-domain-name is RECOMMENDED. An attacker who compromises the DNS for a CA's issuer-domain-name could disrupt validation or potentially impersonate the CA in certain scenarios. While this is a systemic DNS security risk that extends beyond this specification, it is amplified by any mechanism that relies on DNS for identity.
```

are changed to:

```
DNS records are generally not authenticated end-to-end, making them potentially vulnerable to tampering. CAs SHOULD implement additional integrity checks where possible and consider the overall security posture of the DNS infrastructure when relying on persistent validation records.

Additionally, CAs SHOULD protect their ACME Server URL with appropriate security measures. Using DNSSEC to protect the CA's ACME Server URL is RECOMMENDED. An attacker who compromises the DNS for a CA's ACME Server URL could disrupt validation or potentially impersonate the CA in certain scenarios. While this is a systemic DNS security risk that extends beyond this specification, it is amplified by any mechanism that relies on DNS for identity.
```

My reasoning:
The "issuer-domain-name" is a opaque value, that a client should not make requests to.
It could aswell be a non-existent value like:
`4d434f6d36c01ca04fc932a0bb77442fd2471a3ff82d3684798d99ae19f88eea.invalid`

Since the client are not supposed to make requests to, or verify the authenticy of the "issuer-domain-name" - a spoofed or compromised "issuer-domain-name" would not have any security implications - apart from that one CA might fraudulently identify himself as another CA - which would be a violation of CAB/F forum.

Its just a opaque value that are meant to separate for example DigiCert from Letsencrypt from Certum from Google Trust Services.
It does only identify the issuer, it does not AUTHENTICATE the issuer. It could be a digit aswell that marks the CA's index in a public record.

The sensitive part here is the ACME Server URL - which could be fundamentally different from the issuer-domain-name.
If a attacker gains control of the ACME Server URL - the attacker could impersonate the CA, and give the subscriber a "accounturi" he is in control over regardless of which public key the subscriber submits.
And then use that accounturi to issue certificates to the real ACME Server URL.

So I would suggest the wording is changed from "issuer-domain-name" to "ACME Server URL".

sebastiannielsen:
Closes https://github.com/ietf-wg-acme/draft-ietf-acme-dns-persist/issues/63 if merged

This because issuer-domain-name is just a opaque value, and "protecting" that value with DNSSEC - especially if its a legacy domain name that the CA does not own anymore, will not give any security advantage.

It would be analogous to protecting a sign outside a bank with heavy security but have glass doors to the vault.

The sensitive resource for a CA - that needs protecting - is the ACME Directory URL - also called Server URL. Thats the sensitive ACME resource the CA owns and is the root of trust for the ACME client.

### DIFF
--- a/draft-ietf-acme-dns-persist.md
+++ b/draft-ietf-acme-dns-persist.md
@@ -389,7 +389,7 @@ The persistent nature of validation records raises concerns about potential reus
 
 DNS records are generally not authenticated end-to-end, making them potentially vulnerable to tampering. CAs SHOULD implement additional integrity checks where possible and consider the overall security posture of the DNS infrastructure when relying on persistent validation records.
 
-Additionally, CAs SHOULD protect their `issuer-domain-name` with appropriate security measures. Using DNSSEC to protect the CA's `issuer-domain-name` is RECOMMENDED. An attacker who compromises the DNS for a CA's `issuer-domain-name` could disrupt validation or potentially impersonate the CA in certain scenarios. While this is a systemic DNS security risk that extends beyond this specification, it is amplified by any mechanism that relies on DNS for identity.
+Additionally, CAs SHOULD protect their `ACME directory URL` with appropriate security measures. Using DNSSEC to protect the CA's `ACME directory URL` is RECOMMENDED. An attacker who compromises the DNS for a CA's `ACME directory URL` could disrupt validation or potentially impersonate the CA in certain scenarios. While this is a systemic DNS security risk that extends beyond this specification, it is amplified by any mechanism that relies on DNS for identity.
 
 ## Issuer Domain Name Normalization and Limits
 


orangepizza:
doesn't CA/B rule require them to be included in CAAidentifer (from CCADB)?

sebastiannielsen:
> doesn't CA/B rule require them to be included in CAAidentifer (from CCADB)?

Correct, but the value itself is opaque from client's point of view. Meaning, the client doesn't use the value in a particular way - it doesn't fetch a document from the domain, it doesn't verify the domain, it doesn't use the domain at all. 
so lets say letsencrypt would have CAAIdentifier = "letsencrypt.org, freecertificates.org"
and then letsencrypt loses control of freecertificates.org, maybe by losing the EPP transfer code to a malicious party.

That would not mean the security of the ACME protocol is compromised, since nobody uses the domain in the regular sense.
You could aswell have a number that identifies the CA in a public "CA list" or whatever:

```
example.com IN CAA 0 issue "534535;"
example.com IN CAA 0 issuewild "534535;"
```
and
`_validation-persist.example.com IN TXT "534535; accounturi=https://letsencrypt.org/acme/acct/58358493; policy=wildcard"`


The point im making is that the value itself doesn't have any meaning - its just a opaque string. It could be a sha256 hash of a public key or whatever. The value is only used in a string matching to the CAAIdentifier from CCADB.

MEANING: as long as the value exist in CPS and CCADB - its ok. The value itself doesn't need to be secured with DNSSEC. **The CA doesn't even need to own the value.**


The value is only used to capture intent of the domain owner. In the same way voting systems in different countries can be different, some have you write down the number of the president you want to vote on, some have you tick a box at the president's name, some have you write the name of the president, and some have you put pre-filled ballots for your president.

**Since the value doesn't have any meaning in itself, it doesn't matter if a third party owns the value. Even if a CA would call themselves "sebbe.eu" in CAAIdentifier, it would not matter. I would not gain something out of it.
Even if the record would look super weird with:**

`_validation-persist.example.com IN TXT "sebbe.eu; accounturi=http://someca.org/acme/acct/3950359; policy=wildcard"`

**I would still not be able to "compromise" the CA in any way. I would not be able to subvert their ACME traffic, I would not be able to issue for the domain, I would gain nada, zip from a CA "borrowing" my domain as their CAAIdentifier.**

birgelee:
Hi @sebastiannielsen ,

I see your point that there is no constraints on issuer domain name so recommending protection of it may be illogical.

My only thought on your change is is this paragraph needed at all? Recommending protection of ACME directory URL feels a bit like its just something that should have been brought up in the ACME standard.

Just curious if a flat out cut would be a better edit.

I am not particularly opinionated, just wanted to put this out there.

Best,
Henry

MaxHearnden:
It's important to note two things:
1. DNSSEC doesn't prevent DNS based denial of service, it just checks the authenticity of signed records. More records might be required for secure operation than just A/AAAA and without those additional records, an attack such as a BGP hijack might be effective even without spoofing any DNS records.
2. ACME as defined by RFC 8555, while is intended to be transported over HTTPS, is designed to be secure even when unencrypted. The example given for why this security is important is inside a CDN that connects to the backend servers over HTTP. At the moment, dns-persist-01 does rely on HTTPS and so additional security measures might be necessary for secure operation.

sebastiannielsen:
@birgelee Yeah, a cut would work if https://github.com/ietf-wg-acme/draft-ietf-acme-dns-persist/pull/65 is implemented.
But then Pull #65 must be mandatory support.

Since #65 is optional support, I think the paragraph should be included - but correctly be referring to the ACME Server URI.

The security issue mentioned in the paragraph is only relevant for DNS-PERSIST-01 - since that relies on information solely from the ACME server. Not the ACME standard as a whole - since the other authentication methods rely on both a token AND client computed information (mandatory) - which means a malicious ACME server cannot in any way get a client to provision a valid record or file that would allow the malicious party to issue certificates for the domain.

@MaxHearnden 
The problem here is that while the other authentication methods specifically rely on client-computed information AND a token from the ACME server, the DNS-PERSIST-01 method ONLY relies on information from the ACME server. This is the security issue that the DNSSEC paragraph is intended to resolve - if the ACME Server (or its domain) is comrpomised, it means you could issue accounturi's that belongs to an attacker.

Note that even if the ACME protocol is secure and designed to be secure unencrypted, its important that the ACME protocol ONLY focuses on server security. Meaning that the server verfies the client in everything. The client never verifies the server as you can see in the
https://github.com/sebastiannielsen/acme-static
project - I can serve static data to a ACME client and it will happily accept it.
This is however a good thing (as seen in the acme-static project) but that means there must be additional protections against a "malicious ACME server". (attacker-controlled) - when it comes to DNS-PERSIST-01 since it relies solely on ACME-server-provided information.

This is why the security recommendation is only relevant for DNS-PERSIST-01 since that currently relies 100% on server-supplied information.

So I understand fully why the author of the document included that paragraph.
The problem is that the author incorrectly refers to issuer-domain-name, which is a opaque value that has zero meaning in the context.

birgelee:
I am fine going with this change and I see why keeping the language makes sense. Requiring the pub key stuff to be mandatory is something that has been floated but I feel is much larger in scope than this paragraph change.

sebastiannielsen:
@birgelee Now I replied to the question if it should be merged into main or rebased (I choose to merge into main) so now it should be all set to actually approve and merge completely when you want.

There is a discussion making key authorization mandatory for DNS-PERSIST-01 so might be better to wait until that is completed - and actually scrap the portion about DNSSEC and DNS Security if its changed so Key Authorization is always mandatory.

birgelee:
I understand mandatory key authorization is an ongoing discussion, but I think it is safe to move forward with this language change and if we go there we can remove the language in that push.

(FWIW I am against mandatory key authorization and I would like to keep the non-ACME options available so I think this is good language to have in there).

PR #67, opened 2026-07-06
Replace the three accounturi modes with a single hashed account URI

birgelee:
Makes a client-computed hashed URI the sole `accounturi` form, replacing the three modes of -01: the raw ACME account URL (3a), alternative URIs (3b), and subscriber-account URIs (3c). This branch is the working candidate for -02.

The hashed URI is `<accountHashPrefix>/<hash-alg>/<base64url digest>`, where the digest is computed over `length_of_domain || domain_name || key || account_URL`. Binding the account key closes the substitution gap of a raw account URL: a record validates only for an account that controls the key, so a poisoned delivery channel cannot bind a record to an unintended account. Binding the domain gives each domain a distinct record value, which mitigates the correlation risk noted in RFC 8657, Section 5.9. A client that wants one record reused across domains can set `domain_name` to `*` and accept the correlation trade-off.

Changes:

1. Define the hashed-URI construction, with `sha-256` as the initial algorithm and agility through the `<hash-alg>` token.
2. Specify verification by CA-side recomputation, plus the key-retention rules that let records survive routine key rotation. Precedence: account deactivation overrides all retention, then certificate-backed retention, then general retention.
3. Add two directory `meta` fields with IANA registrations: `accountHashPrefix` (base URI for hashed values) and `keyRotationPeriod` (bound on prior-key acceptance).
4. Remove modes 3b and 3c and their dependent text: subscriber-compromise and broad-scope security considerations, DNS-spoofing amplification, mode-specific error handling, and the three-mode revocation table.
5. Rewrite the security considerations around the hashed URI: non-transferability, substitution resistance, enumeration cost, and the stated limits of the privacy protection.
6. Add a worked hashed-URI example that doubles as a test vector, including the domain-omitted `*` form, and update all record examples.


draft-ietf-acme-dns-persist.md
```diff
@@ -163,9 +161,25 @@ The RDATA of this TXT record MUST fulfill the following requirements:
...
+        <accountHashPrefix><base64url hash value>
+        ~~~
+
+        A CA that accepts hashed URIs MUST advertise an `accountHashPrefix` string in the `meta` object of its directory ({{!RFC8555}}, Section 7.1.1). This prefix places hashed account URIs under the CA's infrastructure and lets a client construct them from the directory without a per-provisioning request to the CA. The CA MUST choose a prefix such that appending the base64url hash value yields a valid URI (for example, `https://ca.example/account-hash/`). `<base64url hash value>` is the base64url encoding {{!RFC4648}}, with trailing padding (`=`) omitted, of the SHA-256 {{!RFC6234}} hash digest computed over the following inputs:
+
+        ~~~
+        SHA-256(domain_name || 0x00 || key || account_URL)
+        ~~~
+
+        where `||` denotes octet-string concatenation and `0x00` is a single NUL octet separating `domain_name` from the fields that follow. A NUL octet cannot appear in `domain_name`, so it unambiguously marks the end of that field; `key` is a fixed 32 octets, so the boundary between `key` and `account_URL` is unambiguous as well. The inputs are:
```
MaxHearnden: Unless there's some ACME restriction I'm missing, domain names can actually contain NUL bytes. Domain labels (e.g. www, example and com) are represented in DNS as length + value and can contain any octet including dots, spaces and null bytes (DNS is also case insensitive).

Null bytes cannot appear in base64url encoded data or a URI so the last null byte does unambiguously mark the end of the domain name.

Edit: The key is not base64url encoded and so can contain null bytes. Maybe null bytes in domains should just be prohibited for this method.

CBonnell: > Unless there's some ACME restriction I'm missing

It's not an ACME restriction, but the normalization algorithm prescribed by this spec will fail if the domain name contains ASCII control characters. Specifically, ToASCII will fail as U+0000 is DISALLOWED by RFC 5892.

There's also the RFC 5280 requirement that dNSName GeneralNames be in "preferred name syntax", which would implicitly disallow NUL-containing hostnames to appear in the final certificate.

Given these, the ACME server should reject dns identifiers containing NULs.

birgelee: In general I agree with @CBonnell that both the normalization and the guidance on "preferred name syntax" will both ensure that a domain containing a null byte would not make it this far into the processing (essentially the fact that the hashes are no longer unique is irrelevant as one could not practically get a certificate for this domain anyways via this method).

However, I do tend to prefer standards that are as self-contained as possible and don't rely on restrictions published in other documents. Given that in this encoding the length of the domain name cannot be more than 255 bytes, this single null byte terminator could just be replaced by a length byte leading the domain name. I will probably do that in a commit just to avoid confusion. From my perspective this change has a negligible impact on the implementation.

draft-ietf-acme-dns-persist.md
```diff
@@ -163,9 +161,25 @@ The RDATA of this TXT record MUST fulfill the following requirements:
...
+
+        1. `domain_name` is the FQDN being validated (the Validation Domain Name with its leading `_validation-persist` label removed, no trailing dot), in the normalized lowercase A-label form produced by {{normalization-algorithm}}. This form is US-ASCII.
```
sheurich: Tighten this so it's unambiguous under wildcard/subdomain validation: the hash must bind the FQDN where the record is provisioned (the Validation Domain Name minus the `_validation-persist` label), not the requested certificate FQDN. "the FQDN being validated" invites the wrong reading. Binding the record's own domain is also what neutralizes the wildcard/CNAME uniqueness concern from #64 — worth stating as a positive property here.

draft-ietf-acme-dns-persist.md
```diff
@@ -163,9 +161,25 @@ The RDATA of this TXT record MUST fulfill the following requirements:
...
+    *   **(3b) Hashed URI**: CAs MAY accept hashed URIs. A hashed URI is the concatenation of the CA's `accountHashPrefix` and a base64url hash value:
+
+        ~~~
+        <accountHashPrefix><base64url hash value>
+        ~~~
+
+        A CA that accepts hashed URIs MUST advertise an `accountHashPrefix` string in the `meta` object of its directory ({{!RFC8555}}, Section 7.1.1). This prefix places hashed account URIs under the CA's infrastructure and lets a client construct them from the directory without a per-provisioning request to the CA. The CA MUST choose a prefix such that appending the base64url hash value yields a valid URI (for example, `https://ca.example/account-hash/`). `<base64url hash value>` is the base64url encoding {{!RFC4648}}, with trailing padding (`=`) omitted, of the SHA-256 {{!RFC6234}} hash digest computed over the following inputs:
```
sheurich: The algorithm is fixed to SHA-256 with no agility path. The mitigating factor is real (the CA recomputes rather than a consumer parsing a prefix, unlike the old `pubkey:` proposal), so a CA could migrate by changing `accountHashPrefix` and the algorithm together. Either say that explicitly as the migration story, or add an algorithm identifier. This will come up in WG review given the history.

draft-ietf-acme-dns-persist.md
```diff
@@ -163,9 +161,25 @@ The RDATA of this TXT record MUST fulfill the following requirements:
...
+
+        1. `domain_name` is the FQDN being validated (the Validation Domain Name with its leading `_validation-persist` label removed, no trailing dot), in the normalized lowercase A-label form produced by {{normalization-algorithm}}. This form is US-ASCII.
+        2. `key` is the 32-octet SHA-256 JWK Thumbprint {{!RFC7638}} of a public key associated with the ACME account, as raw digest octets (not base64url-encoded). When provisioning a `_validation-persist` record, the client MUST use the account's current key. A CA MUST accept a hashed URI generated with the current key associated with the account. Additionally, a CA MUST accept a hashed URI generated with a prior key previously associated with the account that is know to the CA per the key hash retention guidance in {{verification-procedure}}. This allows a record provisioned before a key rotation to continue to validate. The rationale for both requirements is given in {{hashed-uri-security}}.
```
sheurich: Two fixes here. First, typo: "that is know to the CA" -> "known." Second, the prior-key acceptance needs to be subordinate to account deactivation — as written, retention would keep a deactivated account's records validating. Add: "A CA MUST NOT accept a hashed URI computed from the keys of a deactivated account, regardless of retention." See the revocation-table comment.

draft-ietf-acme-dns-persist.md
```diff
@@ -207,6 +221,10 @@ _validation-persist.example.com. IN TXT ("authority.example;"
...
+
+A CA SHOULD retain the SHA-256 JWK Thumbprint of every public key an account has used since the CA began accepting `dns-persist-01` records, so that a mode-3b record computed under any of those keys continues to validate. Independent of that guidance, a CA MUST retain the thumbprint of the key used to compute the hashed URI of any mode-3b `_validation-persist` record that most recently validated a domain for a currently valid certificate. Such a thumbprint MUST NOT be pruned while that certificate remains valid, regardless of the key's age, so that a mode-3b record backing an in-use certificate never ceases to validate. Once no currently valid certificate depends on a given key's hashed URI the CA MAY prune that key's thumbprint.
```
sheurich: State this as an explicit CA data-retention obligation, not just a pruning rule: the CA retains {account, key-thumbprint} associations and the link from a currently-valid certificate to the key that validated its backing record. And rank the rules: deactivation overrides the MUST-not-prune-while-cert-valid requirement, so a compromised-key deactivation stops validation immediately even if a valid cert depends on that key.

draft-ietf-acme-dns-persist.md
```diff
@@ -375,63 +393,29 @@ Clients SHOULD protect their ACME account keys with the same level of security a
...
-Section 7.3.5), and the account's key authorization ({{!RFC8555}},
-Section 8.1) provides a verifiable cryptographic binding to the current
-account key at any point in time. Neither property is available for
-modes 3b and 3c, where the URI is CA-assigned and opaque.
-
-Verification options for clients that pre-provisioned records correspond
-to the verifiability gradient above; see {{challenge-object}} for the
-normative requirements. Clients unable to verify SHOULD log the
-substitution for operator review.
+### Account Key Rotation {#account-key-rotation}
```
sheurich, line 422: This PR removed the "Substitution of the Account URI" subsection that used to sit around here, and the challenge-object 3a self-check. Don't drop that content without a replacement: after this change, mode 3a's substitution weakness is both unmitigated and undocumented. At minimum, keep a subsection stating (a) mode 3b is the substitution-resistant path via key binding, and (b) raw mode 3a is a URL commitment only as good as the channel that delivered the account URL, and does not preserve RFC 8555 non-transferability when that URL was poisoned. (If we adopt mandatory binding on the default path, this changes accordingly, but we still need it.)

draft-ietf-acme-dns-persist.md
```diff
@@ -375,63 +393,29 @@ Clients SHOULD protect their ACME account keys with the same level of security a
...
+### Hashed Account URIs (Mode 3b) {#hashed-uri-security}
+
+A hashed URI (mode 3b, {{validation-record-format}}) lets a subscriber commit to an ACME account URL in a `_validation-persist` record without publishing that account URL in cleartext. Its purpose is to reduce the correlation and mass-surveillance exposure described in {{account-uri-privacy}}. It does not hide the account from the CA, which necessarily learns the account URL when it validates the record.
+
+**Source of the authorization binding.** The core of mode 3b's security is the subscriber's commitment to the account URL, as in mode 3a. Because SHA-256 is collision resistant, publishing a digest computed over `account_URL` in a DNS record is a statement of intent to authorize that account that is comparable to publishing the account URL directly.
```
sheurich: Two things for this security subsection. (1) Add a sentence closing Max's length-extension concern on the record: because the digest is compared for exact equality against a CA-recomputed value with fixed field structure, there's no attacker-controlled suffix to exploit, and `domain_name` as the leading field blocks cross-domain linkage. Bare SHA-256 is fine here; HMAC wouldn't help since both sides use only public inputs. (2) The one-to-one binding you describe in the cover note is worth stating precisely here: what RFC 8657 Section 5.4 requires is reverse-direction uniqueness — a URI must not identify two accounts — which holds because `account_URL` is a hash input and SHA-256 is collision resistant. Say "each hashed URI uniquely identifies one account" rather than "1-1 binding."

draft-ietf-acme-dns-persist.md
```diff
@@ -526,12 +510,12 @@ The persistent nature of `dns-persist-01` authorizations means that a valid DNS
...
 3. **Deactivate subscriber account:** For mode 3c only, an immediate deactivation of the entire subscriber account is required. When a subscriber account is deactivated, all associated ACME accounts immediately lose their validation rights derived from the subscriber-account URI. CAs supporting subscriber-account URIs MUST provide this mechanism.
 
 The following table summarizes the applicability and timing of these actions across the three modes:
 
-| Action | 3a (ACME URL) | 3b (Alt URI) | 3c (Subscriber) |
+| Action | 3a (ACME URL) | 3b (Hashed URI) | 3c (Subscriber) |
 |---|---|---|---|
 | Remove DNS record | ✓ (after cache expiry) | ✓ (after cache expiry) | ✓ (after cache expiry) |
 | Deactivate ACME account | ✓ (immediate) | ✓ (immediate) | ✗ (only that account) |
```
sheurich: This "immediate" for 3b is currently unbacked — the normative text that enforced it was removed with the old mode-3b definition. Either restore an equivalent normative rule (see the deactivation comment on the mode-3b definition, line 181) or the table overstates what the spec guarantees. Related: if we cut 3c (see the review body and my mail to the authors), the row below and the 3c column come out entirely, which also resolves the deactivation-semantics gap for that mode — so let's settle the 3c question before reworking this table.

draft-ietf-acme-dns-persist.md
```diff
@@ -571,6 +555,14 @@ IANA is requested to register the following entry in the "Underscored and Global
 - **_NODE NAME**: _validation-persist
 - **Reference**: This document
 
+## ACME Directory Metadata Field {#acme-directory-metadata-field}
+
+IANA is requested to register the following entry in the "Fields in the 'meta' Object within a Directory Object" registry defined in {{!RFC8555}}, Section 9.7.6:
+
+- **Field Name**: accountHashPrefix
```
sheurich, line 580: Changelog flag: -01 was published to the Datatracker with mode 3b meaning "alternative URI." This PR silently repurposes "3b" for the hashed URI. Add an explicit "mode 3b redefined" note in the document, or give the hashed URI a fresh label and retire the old 3b, so implementers tracking -01 aren't caught out. (Also: RFC 6234, cited normatively for SHA-256, is Informational — a downref the IESG will flag at Last Call. Cite FIPS 180-4 or be ready to note the downref.)

sheurich (commented):
Reviewed the hashed-URI construction and the surrounding changes. The core is sound: `SHA-256(domain_name || 0x00 || key || account_URL)`, recomputed by the CA over the requesting account's key, restores RFC 8555 non-transferability for mode 3b. An attacker can't make an honest CA accept a victim's digest for the attacker's account without the victim's private key. The test vector checks out and the privacy framing is honest about being best-effort.

I have three blocking concerns and a set of smaller ones. The largest is that this adds a verifiable mode but leaves the default path (mode 3a, raw account URL) unbound, and deletes the text that documented that gap.

Separately, I've come around to Aaron's position: the binding should be mandatory for the default account path, and mode 3c should be cut from this draft. 3c can't carry the client-computed binding by construction, and the subscriber-account capability survives under BR method 3.2.2.4.22 and CP/CPS without being specified here. That's an authors' decision rather than a line comment, so I've raised it by mail; the inline comments below assume the current mode structure and flag where each item changes if we cut 3c.

Blocking:

1. Mode 3a is still `SHOULD accept` with no client-computed binding, so the substitution attack still succeeds against it, and this PR removes the "Substitution of the Account URI" subsection and the 3a self-check that documented it. We shouldn't ship the deletion without a replacement, whatever we decide on mandatory binding.
2. The revocation table keeps "Deactivate ACME account -> 3b immediate," but the normative text that backed it was removed and the prior-key retention rule would keep a deactivated account's records validating. Deactivation needs to override retention in normative text.
3. The key-retention requirement needs to be stated as an explicit CA data obligation (retain {account, key-thumbprint} plus the cert-to-validating-key link), with deactivation ranked above the MUST-not-prune rule.

Non-blocking: hard-coded SHA-256 with no agility signal; the "3b" label is silently redefined from the published -01 meaning (alternative URI), which is a real changelog hazard since -01 is public on the Datatracker; RFC 6234 is an Informational downref on a standards-track normative reference; and the one-to-one binding claim from the cover note should be stated in the text as the property RFC 8657 Section 5.4 actually needs (each hashed URI uniquely identifies one account, which holds under collision resistance because account_URL is a hash input). Details inline.

draft-ietf-acme-dns-persist.md
MaxHearnden: The URI classification is not in draft-ietf-acme-dns-persist-01. However this would be the first URI in the spec which isn't provided by the server so it might still be note worthy.

draft-ietf-acme-dns-persist.md
MaxHearnden: My concern wasn't with the use of wildcards, it was with the lack of separation between a domain and the ACME account URI which introduced ambiguity when using wildcards. The separator is what mitigates my concern.

MaxHearnden:
> An attacker can't make an honest CA accept a victim's digest for the attacker's account without the victim's private key.

An attacker could still get a client to use their mode 3b (digest) URI as a client's ACME account URI (mode 3a).

birgelee:
@sheurich thanks for the review. Regarding your blocking concerns:
1. My apologies on removing that language. I think I am used to CA/B F documents that contain much less security considerations. Instead of putting the self check back in, it seems people are increasingly in favor of moving towards a system where only 3b is allowed in the draft. I propose doing that in the next update.
2. This was a mistake on my part. Happy to add normative text allowing for deactivate to override acceptance of prior keys.
3. CA data obligation makes sense

other: 
SHA-256 agility: seems to make sense to just put it into the URL like: <ca account URL prefix>sha-256/<hash>
Then alternative algorithms can have cleanly separated URLs. Obviously CAs will need to document which algorithms they are willing to support (either through directory metadata or in CPS)

The change log sounds like a good idea, although I am thinking the next version will not have hashed URLs as "3b". Hashed URLs will just be the primary mode of operating.

RFC 6234 -> FIPS 180-4: Nice catch

1:1 -> "each hashed URI uniquely identifies one account"

birgelee:
@MaxHearnden 

> > An attacker can't make an honest CA accept a victim's digest for the attacker's account without the victim's private key.
> 
> An attacker could still get a client to use their mode 3b (digest) URI as a client's ACME account URI (mode 3a).

True, probably best to just remove 3a.

draft-ietf-acme-dns-persist.md
birgelee: I am still leaning towards going to a length byte at the beginning instead of the separator so the separation is clearly self contained and not dependent on the precondition that there are no null bytes in the domain name.

MaxHearnden: A length byte would also be effective in separating the domain from other fields, [RFC 1035 Section 2.3.4](https://www.rfc-editor.org/info/rfc1035/#section-2.3.4) already limits domain names to 255 bytes.

MaxHearnden:
> SHA-256 agility: seems to make sense to just put it into the URL like: sha-256/

I know this is mostly bikeshedding, however the URL scheme should have support for hash algorithms with a slash in their name, e.g. SHA-512/256 (which can have better performance on 64 bit processors compared to SHA-256).

draft-ietf-acme-dns-persist.md
```diff
@@ -163,9 +161,25 @@ The RDATA of this TXT record MUST fulfill the following requirements:
...
+    *   **(3b) Hashed URI**: CAs MAY accept hashed URIs. A hashed URI is the concatenation of the CA's `accountHashPrefix` and a base64url hash value:
+
+        ~~~
+        <accountHashPrefix><base64url hash value>
+        ~~~
+
+        A CA that accepts hashed URIs MUST advertise an `accountHashPrefix` string in the `meta` object of its directory ({{!RFC8555}}, Section 7.1.1). This prefix places hashed account URIs under the CA's infrastructure and lets a client construct them from the directory without a per-provisioning request to the CA. The CA MUST choose a prefix such that appending the base64url hash value yields a valid URI (for example, `https://ca.example/account-hash/`). `<base64url hash value>` is the base64url encoding {{!RFC4648}}, with trailing padding (`=`) omitted, of the SHA-256 {{!RFC6234}} hash digest computed over the following inputs:
+
+        ~~~
+        SHA-256(domain_name || 0x00 || key || account_URL)
```
sebastiannielsen: I do not think we should use null character as terminating character. In some programming languages this character can be hard to produce.

If a termination character is needed, I would instead suggest a character which are prohibited in a domain, like the pipe character ( |, 0x7c ) - but still can be printed using a regular keyboard, and which is identical in both ISO-8859-1 and UTF-8.

The JWK thumbprint I think should be in base64url - not raw sha256 - most ACME tools already have functions that return a JWK thumbprint in base64url to append to key authorizations, so its benefical if these tools and functions can be reused.

sebastiannielsen:
> moving towards a system where only 3b is allowed in the draft. I propose doing that in the next update.

In that case, I think support for omitting the domain should be added.

My suggestion is then that a client who wishes to opt out of privacy, then  sets domain_name to the wildcard character "*" (0x2a) before hashing.
This will allow a client to opt out of privacy, but still have full security by the other armaments.

A client may wish to provision the same record to multiple domains, or even synthesize using wildcard or use CNAME to point multiple domains to the same validation record.

draft-ietf-acme-dns-persist.md
```diff
@@ -157,39 +164,32 @@ The RDATA of this TXT record MUST fulfill the following requirements:
...
+    3. `key` is the SHA-256 JWK Thumbprint {{!RFC7638}} of a public key associated with the ACME account, as raw digest octets (not base64url-encoded), and is therefore a fixed 32 octets. When provisioning a `_validation-persist` record, the client MUST use the account's current key. A CA MUST accept a hashed URI generated with the current key associated with the account. Additionally, a CA MUST accept a hashed URI generated with a prior key previously associated with the account that is known to the CA per the key retention obligation in {{verification-procedure}}, subject to the deactivation and key-rotation limits stated there. This allows a record provisioned before a key rotation to continue to validate. The rationale for these requirements is given in {{hashed-uri-security}}.
```
sebastiannielsen: Why mot base64url encoded? In this case, it would be preferable, since most ACME libraries already have functions for computing the keyAuthorization for HTTP-01 and DNS-01 authentication, which returns the JWK thumbprint in Base64url encoding.

I feel that base64url should be used here aswell so the same library and functions can be used also for DNS-PERSIST-01.

birgelee: I am personally not that opinionated on this front. I felt it was strange to use a url-friendly encoding in what is a binary input and even if helper functions returned the base64url version it could just be decoded into raw bytes by whatever function was computing this hash.

That said I don't feel strongly about this. I am curious if other WG members have thoughts.

CBonnell: I agree with @sebastiannielsen that using base64url encoding of the thumbprint is easier to implement, because the base64url encoding of the thumbprint is used in other parts of the ACME spec.

draft-ietf-acme-dns-persist.md
```diff
@@ -159,23 +164,32 @@ The RDATA of this TXT record MUST fulfill the following requirements:
...
+    `<hash-alg>` is a lowercase US-ASCII token identifying the hash algorithm used to compute the value. This document defines the token `sha-256` for the SHA-256 algorithm {{FIPS180-4}}. A `<hash-alg>` token MUST NOT contain a "/" octet, so that the "/" following it unambiguously delimits the algorithm identifier from the hash value; an algorithm whose common name contains "/" (for example, SHA-512/256) is represented with a hyphen substituted for the slash. A CA MUST document which hash algorithms it accepts (in directory metadata or in its CP/CPS) and MAY reject an `accounturi` value that uses an algorithm it does not accept. This provides algorithm agility: a CA migrates to a new algorithm by accepting a new `<hash-alg>` token, keeping the algorithm and its identifier bound together in the URI.
```
MaxHearnden: This could require a registry for hash algorithms (if a suitable one doesn't already exist) with algorithm support status (e.g. MUST, RECOMMENDED, MAY, NOT RECOMMENDED, MUST NOT).

draft-ietf-acme-dns-persist.md
```diff
@@ -191,21 +205,38 @@ The RDATA of this TXT record MUST fulfill the following requirements:
...
+
+To verify a hashed-URI `accounturi` value, the CA already knows the requesting account's URL and the FQDN being validated. The CA computes the expected hashed URI as defined in {{validation-record-format}} over each ACME public key it retains for that account (see below); for each such key it computes the value twice, once using the FQDN being validated as `domain_name` and once using the domain-omitted opt-out form (`domain_name` set to `*`). The CA treats the record as satisfying the requirement if its `accounturi` value equals any of these computed hashed URIs under Simple String Comparison ({{!RFC3986}}, Section 6.2.1). Retaining an account's prior key thumbprints allows records provisioned before a key rotation to continue validating.
```
MaxHearnden, line 242: Is there a limit to the number of digests a CA would have to compute i.e. could an account have 200 keys that need to be checked?

It might make sense to either limit the number of concurrently valid keys for validation or have the client tell the ACME server which key is being used.

draft-ietf-acme-dns-persist.md
```diff
@@ -371,67 +402,35 @@ CAs SHOULD implement robust account security measures, including:
...
 
-Domain owners who require privacy without CA cooperation can use separate ACME accounts for domains that should not be correlated. Note that the use of a Subscriber-account URI (mode 3c) represents a privacy inversion compared to the privacy model of mode 3b: a single visible URI correlates all domains across all ACME accounts belonging to that subscriber. Domain owners who require privacy SHOULD use mode 3b or separate ACME accounts, rather than subscriber-account URIs. Domain owners and auditors who require independent verifiability SHOULD use the ACME account URL directly, since third parties cannot independently determine which account is bound to an alternative URI.
+**Limits of the privacy protection.** This construction is a best-effort obfuscation, not a cryptographically strong privacy mechanism. It resists bulk correlation and casual enumeration of account URLs, but a determined adversary with a-priori knowledge of candidate account URLs and their account keys can recover the true account URL. Domain owners who require stronger unlinkability SHOULD use separate ACME accounts for domains that must not be correlated.
```
MaxHearnden, line 452: It might also be worth saying that capturing an http-01 challenge gives the JWK fingerprint of the ACME account which could be enough to perform a brute force attack to recover the account URI.

sebastiannielsen: Its not possible to "capture" an http-01 challenge as that would require knowledge of the token.

Unless the server is configured to respond with a "minted" response based on any token input - but thats a conscious configuration by the server administrator, where the server administator is fully aware of its privacy implications. Its not possible to configure a server in that way by mistake.

HOWEVER, there is a risk when using dns-01 challenges since those are posted at a static location (_validation-persist.DOMAIN.TLD with no random input required), but you should not really mix dns-01 and dns-persist-01 - the server administrator should choose one. So thats a non-issue, if you have access to DNS, there is no reason to mix dns-01 and dns-persist-01, then you can aswell go with dns-persist-01 for everything.

MaxHearnden: Sorry if it wasn't clear, by capture I was referring to something passive on path to the CA.

This concern doesn't apply to dns-01 and tls-alpn-01 as they both use a digest of the key authorisation so the JWK thumbprint can't be recovered.

sebastiannielsen: 1: You can't do this to the ACME client, since the requests between the ACME client and server is encrypted. So even if server administrator sits on a unprotected wifi, it won't matter.

2: The path out from the CA (outgoing validationn requests, unencrypted in http-01) should be assumed to be protected. Its part of the CA's audit that their infrastructure is secure enough.

3: Only way to capture http-01 would then be if the SERVER in question sits on a unprotected wifi or wifi PTP link or if any part of the network on the server side is using hubs instead of switches. Very rare configuration.

4: This would then only have privacy implications, not any security implication.

draft-ietf-acme-dns-persist.md
sebastiannielsen: My suggestion is that the CA should only save one previous key at maximum, and max period to save is 30 days. This makes it clear that if you want to do a key rollover, you have to create the new record.

1 month is enough even for those that don't have access to DNS but need to phone call their web host to have DNS changed,

draft-ietf-acme-dns-persist.md
sebastiannielsen: I think we can do as same as with the ACME account keys - a CA *must* accept sha256, which is the "gold standard" but may accept more. For example, lets encrypt accept alg=ES384 even if thats not part of the standard and normally ES256 is used.

So a registry isn't needed, people coding clients will put in their own registrys in the software for which CAs support which algoritms.

MaxHearnden: JWK algorithm names and semantics are still tracked by an IANA registry https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms.

I couldn't find a similar registry listing hash algorithms and future algorithms could require different semantics e.g. a null algorithm that just base64 encodes the data.

draft-ietf-acme-dns-persist.md
sheurich: Resolved in the direction @sebastiannielsen and @CBonnell describe: draft PR #68 redefines `key` as the 43-octet unpadded base64url RFC 7638 thumbprint, the same encoding ACME key authorizations already use (RFC 8555, Section 8.1), and recomputes both example vectors. The choice is cryptographically neutral and the field stays fixed length, so the deciding factor is that ACME libraries already produce this exact form.

draft-ietf-acme-dns-persist.md
sheurich: The registry you are looking for exists: RFC 6920 established the IANA Named Information Hash Algorithm Registry for exactly this purpose. Draft PR #68 draws `<hash-alg>` from that registry's Hash Name Strings, makes `sha-256` mandatory to implement (the floor @sebastiannielsen describes), and restricts usable entries to non-truncated digests of at least 256 bits, which keeps out truncated variants and anything that is not a real digest. A managed namespace beats per-client tables of CA support: clients and CAs get algorithm agility without pairwise coordination, and each token's semantics stay pinned to its registry entry.

draft-ietf-acme-dns-persist.md
sheurich: Right on the mechanism: HTTP-01 serves `token || "." || base64url(JWK_Thumbprint(accountKey))` over the CA's cleartext validation fetch, so an on-path observer can recover the thumbprint directly; dns-01 and tls-alpn-01 hash the key authorization instead. And agreed on the consequence: with the thumbprint in hand, an attacker can brute-force candidate account URLs offline, so account-URL entropy is the remaining barrier. Draft PR #68 adds exactly this caveat to the privacy analysis: the hashed URI offers best-effort resistance to bulk correlation, not strong unlinkability, and HTTP-01 use reduces the remaining protection to account-URL entropy.

draft-ietf-acme-dns-persist.md
```diff
@@ -159,23 +164,32 @@ The RDATA of this TXT record MUST fulfill the following requirements:
...
+    3. `key` is the SHA-256 JWK Thumbprint {{!RFC7638}} of a public key associated with the ACME account, as raw digest octets (not base64url-encoded), and is therefore a fixed 32 octets. When provisioning a `_validation-persist` record, the client MUST use the account's current key. A CA MUST accept a hashed URI generated with the current key associated with the account. Additionally, a CA MUST accept a hashed URI generated with a prior key previously associated with the account that is known to the CA per the key retention obligation in {{verification-procedure}}, subject to the deactivation and key-rotation limits stated there. This allows a record provisioned before a key rotation to continue to validate. The rationale for these requirements is given in {{hashed-uri-security}}.
```
MaxHearnden: Could the hash algorithm for the JWK thumbprint be the same as the overall hash algorithm e.g. if sha-384 is used to compute the account URI then it should be used for the JWK thumbprint as well.

sheurich: I would keep the two algorithms independent. RFC 8555 Section 8.1 defines the ACME key-authorization thumbprint with SHA-256, and PR #68 deliberately reuses that exact 43-octet base64url value. Making the inner thumbprint follow `<hash-alg>` would define another thumbprint profile for every optional outer hash and prevent clients from reusing ACME's existing component. A stronger outer hash therefore does not remove the construction's SHA-256 dependency, but ACME key authorization has the same dependency. Replacing it should be a coordinated ACME change, not an implicit consequence of selecting a different outer hash.

draft-ietf-acme-dns-persist.md
```diff
@@ -157,25 +179,34 @@ The RDATA of this TXT record MUST fulfill the following requirements:
...
 
-3.  The issue-value MUST contain an `accounturi` parameter whose value is a URI that identifies the ACME account requesting validation, either directly or through a subscriber-account association.
+    ~~~
+    <accountHashPrefix><hash-alg>/<base64url hash value>
+    ~~~
 
-    The `accounturi` value MUST conform to one of the following three modes:
+    A CA that advertises support for the dns-persist-01 challenge type MUST advertise an `accountHashPrefix` string in the `meta` object of its directory ({{!RFC8555}}, Section 7.1.1). This prefix places hashed account URIs under the CA's infrastructure and lets a client construct them from the directory without a per-provisioning request to the CA. The CA MUST choose a prefix such that appending the remaining components yields a valid URI (for example, `https://ca.example/account-hash/`).
```
MaxHearnden, line 190: It might be worth having some provision for multiple valid accountHashPrefixes as might happen if two CAs merge.

sebastiannielsen: > It might be worth having some provision for multiple valid accountHashPrefixes as might happen if two CAs merge.

In fact, the URI does only need to be syntaxially valid. It does not need to be reachable or on the internet.

So a CA could have:

"https://acme.invalid/user-hash/" as account prefix.

The only thing is that if 2 different CA's with separate user databases shares caaIdentities, they need to have distinct URI prefixes.

sebastiannielsen:
since #68 that supercedes this was merged, isn't it time to close this, @birgelee ?

MaxHearnden:
@sebastiannielsen That PR was merged into the content of this PR (it was merged into birgelee-02-candidate) and so this PR is still current.

draft-ietf-acme-dns-persist.md
```diff
@@ -521,21 +523,21 @@ The `persistUntil` parameter provides domain owners with direct control over the
...
+1. **Remove DNS record:** Removing the corresponding DNS TXT record from the Validation Domain Name is the standard method for an Applicant to invalidate a `dns-persist-01` authorization. After the record is removed and resolver caches have expired, new validation attempts for the domain will fail. This action is available to the domain owner and does not require CA involvement.
+2. **Deactivate ACME account:** For situations requiring immediate revocation of issuance capability (e.g., suspected account key compromise), the account bound by the record can be deactivated as specified in {{!RFC8555}}, Section 7.3.6. Per the deactivation rule in {{verification-procedure}}, a CA MUST NOT accept a hashed URI computed from any key of a deactivated account; deactivation therefore takes effect immediately and overrides key retention, including the certificate-backed retention of a key that a currently valid certificate depends on. This makes account deactivation a full emergency measure that invalidates the record's utility without waiting for DNS caches to expire.
```
sebastiannielsen, line 531: I think we should clarify that this MUST only apply to the latest key applied to a deactivated account, not any key in its retention history.

Meaning, if we have three ACME accounts as following:

AccountA with KeyA
AccountB with KeyB
AccountC with KeyX

AccountB performs a keyChange from KeyB to KeyC
AccountA performs a keyChange from KeyA to KeyB
AccountC performs a keyChange from keyX to keyA

Now the accounts looks like this:

AccountA with keyB (retained key: keyA)
AccountB with keyC (retained key: keyB)
AccountC with keyA (retained key: keyX)

AccountA is now "deactivated".


This is now important:
AccountB should now be unable to validate using its old retained keyB (as AccountA = keyB was deactivated).
AccountB should STILL be able to validate using keyC.

But THIS is aswell as important:
AccountC **SHOULD** still be able to validate using keyA (even tough keyA is in the history of the deactivated account AccountA. Meaning, the deactivation **SHOULD** only apply to a account's latest key).
AccountC should also be able to validate using keyX if the CA does perform key retention.


Why is this important, you might ask?

Because, a ACME account MAY change owner for some reason. It could be a person who moves to a new server, where the old account is left with the old server but rekeyed.
It can be a company who changes owner. All keys are replaced for security reasons, but not the accounts itself.

In this case, AccountA may have changed owner, for example, by a company being sold and the new owner takes over the servers.
The new owner of AccountA **may thus not have knowledge of keyA**. Therefore, the new owner of AccountA should thus not be able to cause the deactivation of keyA.

Its important that a person who have previously have had a control over a ACME account, cannot in any way exercise control over that account, including deactivation, by having knowledge of a previous key.

Thats why this deactivation logic should be ONE-WAY - meaning a deactivated account prevents its current key from being used as history, but does not prevent any historic keys from being used as current keys.


This could be worded pretty simple by adding this:

"A CA MUST accept an account's current key for validation regardless of whats previously said, despite any deactivations of any other accounts"

Since two ACME accounts with the same key cannot co-exist, this gurantees that no CAs goes around looking in history of other accounts, while still gurantees that if a account is deactivated, its CURRENT key becomes permanently unuseable.

There MAY be a need for situation where a dns-persist-01 record need to be deactivated even tough the corresponding account has changed the key and theres no current account having that key.
But that is simple - the key owner only needs to create an account using the key, and the immediately deactivate it, to disable the key.

So such a prohibition/requirement does not prevent a key from being deactivated provided the deactivator actually has knowledge of the private key.

draft-ietf-acme-dns-persist.md
```diff
@@ -165,7 +165,23 @@ The RDATA of this TXT record MUST fulfill the following requirements:
 
     *   **(3a) ACME account URL**: CAs SHOULD accept the URI of the ACME account object ({{!RFC8555}}, Section 7.3) as a valid `accounturi` value. A CA that does not accept the account URL directly MUST provide an alternative URI in the challenge `accounturi` field (mode 3b or 3c) and MUST document the accepted `accounturi` format in its Certificate Policy and/or Certification Practice Statement. A CA that advertises support for the dns-persist-01 challenge type MUST accept at least one form of `accounturi` value.
 
-    *   **(3b) Alternative URI**: CAs MAY accept other URIs, provided each such URI uniquely and permanently identifies a single ACME account and satisfies the uniqueness requirements of {{!RFC8657}}, Section 5.4. CAs MUST NOT reassign a URI to a different account. When an ACME account is deactivated, CAs MUST treat all alternative URIs associated with that account as invalid for validation purposes.
+    *   **(3b) Hashed URI**: CAs MAY accept hashed URIs using this schema. A hashed URI is constructed via the following syntax:
+
+    ~~~
+    https://<ACME directory server host domain name>/.well-known/acme-account-hash/<Base64_URL hash value>
```
sebastiannielsen, line 171: I think that we should NOT use .well-known for this.

This because ACME is its own protocol, and does not need to be universially recongnizable. Meaning, I do not think we should register a .well-known.

sebastiannielsen (changes requested):

draft-ietf-acme-dns-persist.md
birgelee: this is an old line and we have since moved away from this. it an ACME directory object now.

Issue #60, opened 2026-04-30
With current verbiage, can't get needed info before issuance


mholt:
I've been implementing DNS-PERSIST in my ACME clients and have hit a snag. I can't get the necessary `issuerDomainNames` except in a Challenge object, which comes from an Authorization, which comes from an Order, because Let's Encrypt, at least, does not support `newAuthz` in its directory.

In other words, DNS-PERSIST essentially makes domain issuance interactive, or an entire Order has to be wasted (which often costs against rate limits).

It should be free to get the needed information to use this challenge. My users will run a command or hit an API endpoint to get the values to put in their DNS records, before any actual certificates are issued. At scale like my clients operate this has to be non-interactive.

I opened [a discussion at Let's Encrypt](https://community.letsencrypt.org/t/dns-persist-without-spending-an-order/246892?u=mholt) since the problem is joint between their implementation of DNS-PERSIST and in the spec, so I'm also opening an issue here, too. Let's Encrypt does not implement `newAuthz` which is an optional endpoint, so I cannot use it to get the `issuerDomainNames`.

While the current CA implementation is the reason I hit a snag, I believe the crux of the problem is the current weak verbiage in the spec which uses "MAY" instead of "MUST". IMO, issue #26 was a step in the wrong direction.

I recommend that the spec either:

- require that `issuerDomainNames` be identical to the `caaIdentities` in the directory (so we can just use `caaIdentities`),
- or revise the spec so that supporting DNS-PERSIST necessarily implies supporting pre-authorization (`newAuthz` then becomes required for DNS-PERSIST)
- or require that `issuerDomainNames` be published to the directory in a new field, although this might be redundant with `caaIdentities`.

Thanks for considering!

(AI-free)

webprofusion-chrisc:
I think publishing (optionally) in the directory meta is the way to go as a signal to acme clients for the _current preferred_ issuer domain name(s). There will be reason not to publish for specific types of CA etc but for the broad case it removes an obstacle. 

I think DNS-PERSIST will be a good addition and I know a lot of my users will deeply appreciate it, if we can remove or de-emphasize very obvious tripping hazards up front it's only good.

commonism:
As you are on the client side, please consider supporting having the apex domain in the challenge as mentioned here: https://github.com/ietf-wg-acme/draft-ietf-acme-dns-persist/issues/33#issuecomment-4302467457.

[...]

Issue #38, opened 2026-02-26
Add client-side persistUntil check guidance


sheurich:
## Problem

The draft places `persistUntil` enforcement entirely on CAs (§4.1 item 5: "CAs MUST NOT consider this validation record valid for new validation attempts after the specified timestamp"). §9.3.1 says CAs SHOULD return `unauthorized` when the timestamp has passed. Nothing tells clients to check `persistUntil` before initiating validation.

A client that manages or provisions `_validation-persist` records knows (or can query) the `persistUntil` value. Without guidance to check it, the client POSTs `{}` to the challenge URL, gets back an `unauthorized` error, and must infer the cause. The remediation (reprovision the record with an updated timestamp) is clear if the client checks locally, but not obvious from the server error alone.

## Proposed change

Add a SHOULD-level recommendation in §9.4 (Client Implementation Guidelines):

> Clients that provision or manage `_validation-persist` records containing a `persistUntil` parameter SHOULD verify that the current time is before the `persistUntil` timestamp prior to responding to a challenge. If the timestamp has passed, the client SHOULD report an actionable error (e.g., "the _validation-persist record for example.com expired on 2026-01-15; reprovision it") rather than initiating a validation attempt that will fail.

Add a forward reference in §4.1 item 5: "See also {{client-implementation-guidelines}} for client-side handling of this parameter."

## Context

Identified by a client implementer working from the draft. The change is SHOULD-level implementation guidance — it doesn't affect the protocol or interoperability. The CA enforces the constraint regardless.

## Open question

Split-horizon DNS complicates client-side checking — enterprise clients may not see the `_validation-persist` record from their network vantage point. Is a SHOULD that silently fails behind split-horizon DNS worth adding?

sheurich:
@rmbolger Good point on split-horizon. The SHOULD doesn't require a DNS query; a client that provisioned the record already has the `persistUntil` value in its own state. The check is local: compare the current time against the timestamp the client chose when it wrote the record. Split-horizon DNS is irrelevant because the client never resolves the validation record to perform this check.

For the out-of-band case you're describing (an enterprise workflow provisions the record and a separate ACME client just triggers validation) the client genuinely doesn't know `persistUntil`. That's fine. The SHOULD naturally doesn't apply; the CA enforces the constraint regardless, and the client gets an `unauthorized` error it can surface to the operator. This matches the LE forum pattern where clients don't manage DNS records.

[...]

PR #68, opened 2026-07-21
Apply review fixes and issue resolutions (#38, #42, #56, #60)

sheurich:
Fixes against the -02 candidate (#67), reviewed at f816113. Both test vectors recomputed; build and lint clean.

Review fixes:

- Rewrite the stale Account Key Rotation text: the account URL is stable, but the hashed accounturi changes with the account key, so clients republish after rollover.
- Clarify that the CAA accounturi (the account URL, RFC 8657 Section 3.1) and the dns-persist accounturi (the hashed URI) are distinct values matched by different procedures.
- Document HTTP-01 thumbprint exposure as a privacy limit.

Construction changes from the #67 review threads:

- Hash the 43-octet unpadded base64url RFC 7638 thumbprint, the same form as the key-authorization component.
- Define hash-alg as an RFC 6920 Hash Name String; sha-256 mandatory for clients and CAs.

Issue resolutions:

- #38: clients check persistUntil against local provisioning state before starting an order.
- #42: DNS TTL no longer limits validation-data reuse; reuse is bounded by persistUntil and the CA's Validation Data Reuse Period.
- #56: the challenge-object member issuer-domain-names becomes issuerDomainNames; DNS parameter names are unchanged.
- #60: CAs must advertise issuerDomainNames in directory metadata, with the D ⊆ C consistency rule and the IANA registration. C ⊆ I applies only when caaIdentities is advertised.

Two additions go beyond what the issue threads literally agreed; let me know if you disagree:

- Optional hash algorithms must be 256-bit or longer, non-truncated registry entries (the RFC 6920 registry includes sha-256-32).
- persistUntil now also ends reuse of previously obtained validation data; the old record-format sentence contradicted the #42 direction.

Deliberately untouched: keyRotationPeriod and the domain-omitted * form; those will be follow-ups.

Refs #38, #42, #56, #60.

### DIFF
--- a/draft-ietf-acme-dns-persist.md
+++ b/draft-ietf-acme-dns-persist.md
@@ -153,15 +153,17 @@ A CA offering the `dns-persist-01` challenge type MUST advertise an `issuerDomai
 
 The directory `issuerDomainNames` array MUST use the same normalization and length rules as the challenge object's `issuerDomainNames` array ({{challenge-object}}): each entry MUST be a lowercase A-label domain name with no trailing dot and MUST NOT exceed 253 octets. The array MUST NOT be empty and MUST NOT contain more than 10 entries.
 
-Let D be the set of values in the directory `issuerDomainNames` array, and let I be the set of values in the directory `caaIdentities` array ({{!RFC8555}}, Section 7.1.1), represented as lowercase A-label domain names with no trailing dot. A CA offering `dns-persist-01` MUST advertise `caaIdentities`. For each `dns-persist-01` challenge object the CA issues, let C be the set of values in that challenge's `issuerDomainNames` array. The CA MUST satisfy these requirements:
+Let D be the set of values in the directory `issuerDomainNames` array, and let I be the set of values in the directory `caaIdentities` array ({{!RFC8555}}, Section 7.1.1), represented as lowercase A-label domain names with no trailing dot. This document does not require a CA to advertise `caaIdentities`; a CA that does not perform CAA validation has no names to disclose there. For each `dns-persist-01` challenge object the CA issues, let C be the set of values in that challenge's `issuerDomainNames` array. The CA MUST satisfy these requirements:
 
 - Every value in D MUST appear in C. A client that pre-provisions a record using a name from D is therefore guaranteed that the name will satisfy the `issuer-domain-name` requirement of any `dns-persist-01` challenge that CA issues later, without first opening an authorization.
 - C MAY include values beyond D while remaining within the 10-entry maximum in {{challenge-object}}, so a CA can offer challenge-specific names in addition to its pre-provisioning baseline.
-- Every value in C MUST also appear in I. This prevents a CA from accepting an `issuer-domain-name` that it has not also disclosed through `caaIdentities`.
+- If the CA advertises `caaIdentities`, every value in C MUST also appear in I. This prevents a CA that participates in CAA validation from accepting an `issuer-domain-name` it has not also disclosed through `caaIdentities`.
 
-For these subset comparisons, the CA MUST compare `caaIdentities` values after applying the same lowercase A-label and trailing-dot rules used for D and C.
+For these subset comparisons, the CA MUST compare `caaIdentities` values, when advertised, after applying the same lowercase A-label and trailing-dot rules used for D and C.
 
-A client MUST treat a missing or nonconforming directory `issuerDomainNames` or `caaIdentities` array as unavailable for pre-provisioning. It MAY still process an interactive `dns-persist-01` challenge. When D is available, the client MUST consider a challenge malformed if any value in D is absent from C. When I is available, the client MUST consider a challenge malformed if any value in C is absent from I.
+`caaIdentities` values of publicly trusted CAs are kept distinct by the CA/Browser Forum disclosure and audit regime. Identities of CAs outside that regime have no equivalent protection against another CA recognizing the same name, so a domain owner who adds such an identity to a CAA policy accepts that reuse caveat.
+
+A client MUST treat a missing or nonconforming directory `issuerDomainNames` array as unavailable for pre-provisioning. A missing or nonconforming `caaIdentities` array does not affect pre-provisioning; it only makes the I comparisons unavailable. A client MAY still process an interactive `dns-persist-01` challenge. When D is available, the client MUST consider a challenge malformed if any value in D is absent from C. When I is available, the client MUST consider a challenge malformed if any value in C is absent from I.
 
 # Challenge Response and Verification {#challenge-response-and-verification}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-acme-dns-persist.md
```diff
@@ -143,11 +142,27 @@ The following shows an example challenge object:
...
+
+Let D be the set of values in the directory `issuerDomainNames` array, and let I be the set of values in the directory `caaIdentities` array ({{!RFC8555}}, Section 7.1.1), represented as lowercase A-label domain names with no trailing dot. A CA offering `dns-persist-01` MUST advertise `caaIdentities`. For each `dns-persist-01` challenge object the CA issues, let C be the set of values in that challenge's `issuerDomainNames` array. The CA MUST satisfy these requirements:
+
+- Every value in D MUST appear in C. A client that pre-provisions a record using a name from D is therefore guaranteed that the name will satisfy the `issuer-domain-name` requirement of any `dns-persist-01` challenge that CA issues later, without first opening an authorization.
+- C MAY include values beyond D while remaining within the 10-entry maximum in {{challenge-object}}, so a CA can offer challenge-specific names in addition to its pre-provisioning baseline.
+- Every value in C MUST also appear in I. This prevents a CA from accepting an `issuer-domain-name` that it has not also disclosed through `caaIdentities`.
```
MaxHearnden: Requiring CAA for this method could pose a risk for organisations using private CAs as the organisation could have to publish additional CAA records where the CAA identity is not protected from reuse by another CA.

This might be better suited for the CA/B Forum TLS baseline requirements.

draft-ietf-acme-dns-persist.md
```diff
@@ -162,7 +177,7 @@ The RDATA of this TXT record MUST fulfill the following requirements:
 
 1.  The RDATA value MUST conform to the issue-value syntax defined in {{!RFC8659}}, Section 4.2. To ensure forward compatibility, the server MUST ignore any parameter within the issue-value that has an unrecognized tag.
 
-2.  The `issuer-domain-name` portion of the issue-value MUST be one of the Issuer Domain Names provided by the CA in the `issuer-domain-names` array of the challenge object. If the `issuer-domain-name` does not match any of the provided values, the CA MUST reject the record.
+2.  The `issuer-domain-name` portion of the issue-value MUST be one of the Issuer Domain Names provided by the CA in the `issuerDomainNames` array of the challenge object. If the `issuer-domain-name` does not match any of the provided values, the CA MUST reject the record.
 
 3.  The issue-value MUST contain an `accounturi` parameter whose value is a hashed URI identifying the ACME account requesting validation. A hashed URI cryptographically binds the account to the domain being validated without publishing the account URL in cleartext. It is formed by concatenating the CA's `accountHashPrefix`, a hash-algorithm identifier, a "/" separator, and a base64url hash value:
```
MaxHearnden, line 184: Using the Named Information Hash Algorithm Registry might mean that we should use one of the formats defined by RFC 6920 (https://datatracker.ietf.org/doc/html/rfc6920), in particular we could use the URL segment format (e.g. sha-256;UyaQV-Ev4rdLoHyJJWCi11OHfrYv9E1aGQAlMO2X_-Q) or the .well-known URL format (e.g. http://example.com/.well-known/ni/sha-256/UyaQV-Ev4rdLoHyJJWCi11OHfrYv9E1aGQAlMO2X_-Q).

draft-ietf-acme-dns-persist.md
sheurich: Fair catch. Two fixes coming out of this.

The intent of C ⊆ I is a single identity vocabulary: the name in a domain's CAA policy and in its `_validation-persist` record should be the same string for the same CA. That only makes sense where the CA participates in CAA validation, so the fix is scoping rather than dropping: the blanket "MUST advertise `caaIdentities`" is removed and the rule becomes conditional. If the CA advertises `caaIdentities`, every value in C MUST also appear there. A CA that doesn't do CAA validation never touches the CAA identity namespace, and a new sentence notes the identity-reuse caveat for CAs outside the CCADB/BR disclosure regime.

Your comment also surfaced an adjacent bug: as written, a missing `caaIdentities` array knocked out pre-provisioning entirely. That should only depend on `issuerDomainNames`; fixed in the same commit.

For publicly trusted CAs the mandate has audit teeth in the BRs rather than in this document; that side belongs with the CABF Method 3.2.2.4.22 cleanup work (cabforum/servercert#674-#677 territory).

Fixed in 5680afd.

MaxHearnden:
> Optional hash algorithms must be 256-bit or longer, non-truncated registry entries (the RFC 6920 registry includes sha-256-32).

This will make sure that we're [SP 800-131A Rev. 3](https://csrc.nist.gov/pubs/sp/800/131/a/r3/ipd) (currently a draft) compliant (migrating from 112 to 128 bits of security).

draft-ietf-acme-dns-persist.md
sheurich: The narrow reuse is intentional: this draft uses the RFC 6920 Hash Name String registry, not the RFC 6920 name formats. The value here identifies an ACME account under the CA's `accountHashPrefix`; it does not name retrievable content whose bytes are checked against the digest. Requiring an `ni:` URI or the `/.well-known/ni/` HTTP mapping would import content-naming and dereferencing semantics this value does not have, and it would constrain the CA's URI layout. The draft therefore does not adopt or depend on either format.

birgelee:
I am all good on merging this. One one comment is on the new proposal to have DNS persist records break on ACME key change. I know the records are computed over the key, but I think there is a strong security justification for why they don't need to break on key rollover. I will probably chat more about this coming up. I like to keep everything moving so I would rather merge this in and then change the behavior if we decide to go there.

sebastiannielsen:
> I am all good on merging this. One one comment is on the new proposal to have DNS persist records break on ACME key change.

I feel that if someone wants to do key rollover, they should also replace the record. Keeping a old record indefinitely creates security risks, especially if a key is moved from an account to another - which creates ambigiutiy, especially if 2 CAs share infrastructure, have different caaIdentities, but identical accounturi's and account databases.

(the requirement for 2 different CAs to have distinct account uri's only applies if they also share caaIdentities).

The prescribed keyRotationPeriod meta record allows a client to do a emergency key rotation, and then do the record replacement.

Which can be done like how DNSSEC key rotation is made - publish the new record, wait until it goes live in all DNS caches, and only after then, delete the old record.

I don't see a need for supporting infinitely old records just to allow key rotation with no access to DNS at all, which increases storage space and requirements for CA's. Its enough they need to retain all deactivated accounts, having to retain all keys - especially now when ML-DSA is waiting behind the door will create problems, especially if a client does multiple key rotations.

Issue #69, opened 2026-07-26
Why permitting DNS persist records after key compromise and rollover still hold security properties


birgelee:
In the IETF meeting and on list there were several discussions about whether a DNS-persist DNS TXT record which now has the hash of the ACME public key in it should be allowed to continue validating a domain after that ACME key has been rotated. I have some appreciation for the general sentiment that a record made on a stale (and possibly compromised key) should not continue to validate a domain. However, I want to open this issue to ***present the reasoning that old DNS-persist records could continue to be validated post key rotation while upholding the security model of DNS persist***. This even includes resilience to the ACME proxy attack which is now included in our threat model for this record.

This reasoning is based on two security assumptions:

1. A currently valid key for an ACME account is not compromised. Thus, until the moment the subscriber requests a key rollover, that key is valid. Obviously after rollover the old key must be assumed to be compromised and even potentially assigned to an adversary ACME account. While obviously there is always a risk of ACME account key compromise, the security of current ACME account keys has always been a part of the ACME protocol as DCV objects can be reused.
2. A subscriber's account URL properly identifies the true subscriber even after key rollover. Essentially this property comes down to the principle that the subscriber still maintains access to their true account after rollover. Even after rollover, the account (identified with the Account URL from the location header) is still controlled by the victim.

To justify why records made with the scheme in the current v2 candidate branch can continue to securely authorize an ACME account post key rollover (under particular conditions I will outline) its best to break the DNS-persist validation into two steps: initial validation (the first time a domain is validated with a DNS persist record) and persistent validation (when a certificate is renewed using an old record that was already used to validate that domain previously). If we can prove security of the record in these two conditions, separately, the record has security overall.

Additionally I will place two conditions on the acceptance of old records which will be used in this security analysis. condition 1: a DNS persist record that is being validated for the first time MUST contain the hash of the CURRENT ACME account key. Historical keys are not accepted on initial validation.
condition 2: during persistent validation, only a record that has the same hashed account URI which previously passed initial validation can be used. Note that the key that was original contained in that hash may be rolled, but at the time it was made it must have passed initial validation.

Now to go through the security reasoning.

[...]

PR #70, opened 2026-08-02
Implements #69 (Hashed URI retention instead of key retention)

sebastiannielsen:
This implements #69 by changing key retention into retaining previously validated URI hashes instead.

This however removes the ability for the CA from preventing issuances for a disabled key (that have been rotated out) since the CA cannot reverse a stored hash.

This is compensated for by allowing a certificate keypair based revocation, to also prune all previously used hashed URIs, which allows a account owner of a compromised ACME account, to prevent issuance using that account, even if the ACME account owner doesn't temporarly not have access to DNS to remove the record, by requesting a certificate keypair-based revocation.


sebastiannielsen:
Closes #69 if implemented.

draft-ietf-acme-dns-persist.md
```diff
@@ -239,21 +239,22 @@ _validation-persist.example.com. IN TXT ("authority.example;"
...
 
-To recompute these hashed URIs, a CA has a data-retention obligation: it MUST retain an association between each ACME account and the SHA-256 JWK Thumbprints of the public keys that account has used, and it MUST retain, for each currently valid certificate, the link from that certificate to the specific key that computed the hashed URI of the `_validation-persist` record that most recently validated a domain for it. The following rules govern this retained data and the acceptance of the keys behind it, listed in order of precedence with earlier rules overriding later ones:
+To reuse these hashed URIs, a CA has a data-retention obligation: it MUST retain an association between each ACME account and the sha256 hashes that have passed initial DNS-PERSIST-01 validation along with for which domain. The following rules govern this retained data and the acceptance of the keys behind it, listed in order of precedence with earlier rules overriding later ones:
```
MaxHearnden, line 244: Multiple hash algorithms can be used to calculate the account URI not just SHA-256, it might be better to refer to the hashed account URI or since the retained URI appears to only be stored for comparison with the retrieved account URIs, a digest of the hash account URI (not the ACME account URI).

MaxHearnden (changes requested):

draft-ietf-acme-dns-persist.md
```diff
@@ -239,21 +239,22 @@ _validation-persist.example.com. IN TXT ("authority.example;"
...
+To verify a hashed-URI `accounturi` value, the CA already knows the requesting account's URL and the FQDN being validated. The CA computes the expected hashed URI as defined in {{validation-record-format}} over each ACME public key it retains for that account (see below); for each such key it computes the value twice, once using the FQDN being validated as `domain_name` and once using the domain-omitted opt-out form (`domain_name` set to `*`). The CA treats the record as satisfying the requirement if its `accounturi` value equals any of these computed hashed URIs under Simple String Comparison ({{!RFC3986}}, Section 6.2.1). Retaining an account's hashed URIs that previously passed validation allows records provisioned before a key rotation to continue validating.
```
sheurich, line 242: This still recomputes over retained account keys, but the next paragraph removes key retention. Define first validation and steady-state reuse separately: a new value must match the current key or another key the account verifiably held; a previously accepted value may match the exact stored `accounturi` for the same `(account, validation domain)`. A successful current-key validation should refresh the stored value.

draft-ietf-acme-dns-persist.md
sheurich, line 244: I agree with Max that this must retain the exact algorithm-tagged `accounturi`, not a "sha256 hash." The storage key also needs to be `(account, validation domain)` so deactivation and domain scoping do not require reversing anything.

draft-ietf-acme-dns-persist.md
```diff
@@ -239,21 +239,22 @@ _validation-persist.example.com. IN TXT ("authority.example;"
...
-3. **General retention.** Subject to the rules above, a CA SHOULD retain the thumbprint of every public key an account has used since the CA began accepting `dns-persist-01` records, so that a record computed under any of those keys continues to validate. A CA MAY bound acceptance of a rotated-out key with a `keyRotationPeriod` (see {{key-rotation-period}}); once no currently valid certificate depends on a given key's hashed URI, and any applicable `keyRotationPeriod` has elapsed, the CA MAY prune that key's thumbprint.
+1. **Certificate key-revocation overrides retention.** A CA MUST NOT accept a hashed URI that was successfully passed initial validation BEFORE a revocation using the certificate key as outlined in ({{!RFC8555}}, Section 7.6) using the certificate keypair. The CA MAY limit this operation to certificates issued using DNS-PERSIST-01, if the CA has this information stored. After such a revocation is made, the CA must prune any previously validated hashed URIs from the account.
```
sheurich, line 246: Account deactivation must remain authoritative. Because stored values belong to an account record, deactivating that account can purge its entries directly; the CA does not need to reverse the digest or scan other accounts. Certificate-key revocation is not a substitute for this rule.

draft-ietf-acme-dns-persist.md
```diff
@@ -239,21 +239,22 @@ _validation-persist.example.com. IN TXT ("authority.example;"
...
+2. **General retention.** Subject to the rules above, a CA SHOULD retain the hashed URI of every passed DNS-PERSIST-01 validation passed on a particular account, so that a record continues to validate even after key rotation. A CA MAY bound acceptance of a rotated-out key with a `keyRotationPeriod` (see {{key-rotation-period}}); once any applicable `keyRotationPeriod` has elapsed or a certificate was revoked using the certificate keypair as outlined in 1, the CA MAY prune that hashed UEI.
+
+An CA MUST recheck that a stored hashed URI still is present in DNS - The stored hashed URIs only serve as comparision strings to allow an account with a rotated out key to still validate using a previously published DNS record - AS LONG as that DNS record have previously passed validation at least once.
```
sheurich, line 249: Presence of the stored value is not enough. Every attempt still needs the normal DNS lookup and complete record evaluation, including issuer, policy, `persistUntil`, and CAA. Stored state changes only the account-binding comparison.

draft-ietf-acme-dns-persist.md
```diff
@@ -239,21 +239,22 @@ _validation-persist.example.com. IN TXT ("authority.example;"
...
 
 ## Account Key Rotation Period {#key-rotation-period}
 
-A CA MAY advertise a `keyRotationPeriod` in the `meta` object of its directory ({{!RFC8555}}, Section 7.1.1). When present, its value MUST be a non-negative integer number of seconds. It bounds how long, after a public key ceases to be the current key of an ACME account (for example, because the account rotated to a new key per {{!RFC8555}}, Section 7.3.5), the CA continues to accept hashed URIs computed with that prior key.
+A CA MAY advertise a `keyRotationPeriod` in the `meta` object of its directory ({{!RFC8555}}, Section 7.1.1). When present, its value MUST be a non-negative integer number of seconds. It bounds how long, after a public key ceases to be the current key of an ACME account (for example, because the account rotated to a new key per {{!RFC8555}}, Section 7.3.5), the CA continues to accept old hashes that was previously validated with that key.
```
sheurich, line 253: `keyRotationPeriod` is no longer the editor proposal. The current direction is non-breaking validated-value retention, bounded by DNS removal, `persistUntil`, deactivation of the requesting account, and the agreed certificate-lifetime lapse. Please remove this section rather than adapting the old window.

draft-ietf-acme-dns-persist.md
```diff
@@ -447,7 +448,7 @@ The hashed URI is the sole `accounturi` form ({{validation-record-format}}). It
...
+**Acceptance of prior keys.** A CA accepts digests that was previously accepted so that records provisioned before a rotation survive without republication ({{!RFC8555}}, Section 7.3.5), subject to the certificate-key-based revocation, and key-rotation-period rules in {{verification-procedure}} and {{key-rotation-period}}. Such a record was committed to the account's URL when it was provisioned, so even if the key in the record is one that was later rotated out, the domain owner's intent to authorize that account is preserved. Certificate key-pair revocation prunes the previously accepted hashed URI list of an account, which is the emergency stop for a suspected key compromise.
```
sheurich, line 451: Prior successful validation does not prove the domain owner's intent to authorize this account under the planted-account attack. The supported claim is narrower: the stored value preserves CA-local continuity for the same account object and validation domain. Authenticated account-resource proof is now the author proposal for strengthening creation integrity, pending implementation details and co-author/WG review.

draft-ietf-acme-dns-persist.md
```diff
@@ -528,16 +529,18 @@ The persistent nature of `dns-persist-01` authorizations means that a valid DNS
...
+2. **Deactivate ACME account:** For situations requiring immediate revocation of issuance capability (e.g., suspected account key compromise), the account bound by the record can be deactivated as specified in {{!RFC8555}}, Section 7.3.6. Note that this does not prevent issuance if a ACME account that previously have validated a hashed URI, becomes compromised and the attacker keyChanges into a key the attacker controls. For this situation, either 1, and/or 3, must be carried out to cease acceptance of the comrpomised key.
+3. **Revoke a certificate using its keypair:** For situations where the account holder has lost control of a compromised ACME account, the account owner can request revocation of a certificate issued using DNS-PERSIST-01, using the account, using the certificate keypair as specified in as specified in {{!RFC8555}}, Section 7.6. This clears all previously accepted hashed URIs on the compromised account. After that, the account must perform a ACME account deactivation using the compromised keypair, to prevent the attacker from keyChanging back to the compromised account key to aquire a fresh validation using the old record.
```
sheurich, line 533: RFC 8555 Section 7.6 authorizes revocation of the named certificate; it does not authorize deleting every persistent value for the issuing account. A certificate key may be held by a delegated server with no authority over the account's other domains, so this creates an account-wide denial-of-service path. Certificate-key-triggered pruning is deferred until after `-02`; any later proposal needs a separate authority, scope, and threat-model decision.

sheurich (changes requested):
Storing the exact value that passed validation is the right steady-state direction, but this patch should not merge as written. It retains the old key-enumeration algorithm while removing key storage, stores an underspecified digest rather than the complete `accounturi`, loses first use of a record provisioned before rollover, removes account-deactivation precedence and the certificate-lifetime lapse, keeps `keyRotationPeriod`, and gives certificate-key revocation a new account-wide side effect.

The author patch should define first validation separately from exact stored-value comparison per `(account, validation domain)`, keep full DNS evaluation, DNS removal, `persistUntil`, and account deactivation authoritative, and omit certificate-key-triggered pruning from `-02`.

draft-ietf-acme-dns-persist.md
sebastiannielsen: Oooh I didnt notice that in the text. Ill change it. But lets discuss the other changes first.

draft-ietf-acme-dns-persist.md
sebastiannielsen: Exactly what 249 says. That the CA must recheck DNS (still present in DNS), it can't just store the values and check against a local database to authorize the domain without asking DNS at all.

Meaning, the stored "approved" authorizations does not replace the normal authorization procedure. It just authorizes the CA to accept a stored value that was previously approved even if the hash no longer computes.

draft-ietf-acme-dns-persist.md
sebastiannielsen: Correct, the storage key must be account and domain - in this way, a prior accepted authorization for lets say example.com would not authorize for example.org even if they have a identical record in DNS.

Meaning, each prior acceptance gets "locked" to the domain basically.


Example:
Lets say I compute a record using *, my account key, and my account uri.
Now I can provision this record to any domain.

If I now provision the record to example.com and example.org

Then I validate example.com

Then I keyChange my account to a new key.

The result of my proposal means, that example.com will validate (since that record validated at least once) but example.org will refuse to validate.

draft-ietf-acme-dns-persist.md
sebastiannielsen: Here is the problem:

The previous deactivation would deactivate a stored key - cross account.
If the rule really meant only keys on the very same account and that the CA should not scan for deactivated account cross-account, then the rule is completely unneccessary since a deactivated account is unuseable - it can't create orders, it can't fetch certificates, it can't create authorizations, it can't submit challenges, nothing.

So for example:
If AcctA with keyA keychanges into keyB.
Then acctB is created with keyA

if acctB would now be deactivated:
Then acctA would need to be unable to validate using a stored record for the prior keyA.

And I understand why thats important: If AcctA is comrpomised, and attacker changes key into keyB, the real owner is unable to access the account. So this is why the key-deactivation mechanism is important -> The real owner can then create an account using keyA, and then immediately deactivate it, to rob AcctA of its issuance rights for the domain from which the account was stolen from.


Since the new system now stores HASHES, it would be unable to check for key-usage cross-account - since that would require reversing the hash.

But I think there need to be a system, where a account owner who got his account stolen, can "nuke" the prior validations without having access to DNS so he could delete the record. And here is where certificate keypair revocation comes into play.


Note that the rule was requiring the CA to scan the key thumbprint for cross-account deactivation.
Today, an deactivated account CANNOT be used to issue certificates at all, so a "deactivation takes precendence" rules isn't needed if you aren't out for a cross-account scan.

draft-ietf-acme-dns-persist.md
sebastiannielsen: Could you explain more what you mean with the "planted account attack"?

draft-ietf-acme-dns-persist.md
sebastiannielsen: Yeah, I agree about that, but thats the only mechanism for which an account could "revoke" an issuing account possibility to gain new certificates using old stored URI hashes if the account owner lost control of the account.

One way would be to limit this type of revocation to only the domains present in the certificate being revoked, so it would render only those stored "key rotated" URI hashes deleted?

draft-ietf-acme-dns-persist.md
sebastiannielsen: I agree with a better idea would be to store the whole account URI. Or even better, the whole DNS record (the whole DNS TXT resource).

So a record can only survive if its unchanged from a keyrotation. Once it being changed, the record MUST be updated to the new key. That should make any attacks on the system much harder.

draft-ietf-acme-dns-persist.md
sheurich: Shorthand for the pre-registered-account substitution Max raised on #64 and #69. Since the record's accounturi is an opaque hash, the domain owner can't determine which account an unfamiliar value names; recomputation only confirms a guess. Two ways such a value reaches DNS. One, someone is persuaded to publish the attacker's complete precomputed value (a hosting or integrator flow that says "add this TXT record to enable certificates"); that record identifies the attacker's account at the target CA immediately. Two, an ACME intermediary hands the client a foreign account URL, and the client publishes H(domain, victim key, attacker URL). That record matches no account today, but it isn't dead: the keyChange duplicate-key check (RFC 8555 7.3.5) only refuses keys currently held by another account, so once the victim rotates away from that key and it later leaks, the attacker can bind the leaked key to their pre-registered account and the planted record validates for them. (If the victim never held an account at that CA, even the rotation step is unnecessary; the leak alone suffices.) Rotation-then-compromise is exactly the window this thread is about.

That's why prior successful validation can't prove the domain owner meant to authorize the account. It proves only that this account matched this record at that time. Storing the value therefore supports CA-local continuity for the same account and validation domain, and nothing stronger.

On the creation side, the author proposal (pending co-author/WG review) is that the client prove its key authenticates to the exact account URL used in the digest (POST-as-GET) before publishing. That checks the key and URL in the digest belong to one account the client controls, which forecloses the second path before anything reaches DNS. It doesn't help when someone pastes a complete precomputed value straight into DNS, it doesn't stop an attacker who writes the record directly during a DNS compromise, and a proxy that terminates the client's TLS toward the CA can fake the proof, so those belong in Security Considerations as residual risks.

sebastiannielsen: My suggestion to solve this problem, is to use my proposal in the comment on the issue:
Store the whole TXT record, and ONLY accept that for the PARTICULAR account.

So when a account A validates a DNS-PERSIST-01 record, the CA will check:

does the TXT record EXACTLY match a previously stored TXT record for this account and for this domain?
Then skip the accounturi check (no digesting) - and only check persistUntil, MPIC and CAA. (the issuer-domain doesn't need to be checked since thats part of the stored value).

Then authorize the issuance.

If it doesn't exist - THEN:
Digest domain, accounturi and JWK thumbprint. Does it match the hash? YES then authorize the issuance (provided persistUntil, issuer-domain, MPIC and CAA checks out)
STORE the complete TXT record for the account and domain. (meaning account AND domain should be key)

If it doesn't match, then:
Digest "*", accounturi and JWK thumbprint. Does it match the hash? YES, then authorize the issuance (provided perssistUntil, issuer-domain, MPIC and CAA checks out)
STORE the complete TXT record for the account and domain. (meaning account AND domain should be key)

(Note here that the domain should be stored even if its an "\*" record - why you might ask - this LOCKS the stored record to the previously authorized domain - so even if its an "\*" record, it CANT be moved to another domain UNLESS the hash is for a current key of the account - MEANING any "old keys" need to have a previous issued certificate to be able to "remain" in DNS unchanged over a keyChange)

sebastiannielsen: Im TOTALLY against a extra request for "verifying" a accounturi. Why?

Because, to be effective, the check must be made mandatory - for example by adding a token value that must be inserted into DNS. (which defeats the purpose of a "PERSIST")

And it also wastes resources by additional requests.

sebastiannielsen: Note that the mechanism (with disabling "concurrent" deactivated keys cross-account) you are describing would not prevent the H(domain, victim key, attacker URL) attack, since the victim key does not belong to a deactivated account. The attacker has control of the victim key as soon as it leaks.

Only way this would be prevented is blacklisting old keys after each keychange, so keyChanged keys effectively becomes attached to a "new virtual" account that is automatically deactivated.

sheurich: > My suggestion to solve this problem, is to use my proposal in the comment on the issue: Store the whole TXT record, and ONLY accept that for the PARTICULAR account.
> …

Thanks, I understand the whole-record algorithm now. On a stored exact match, the CA can skip recomputing accounturi, but it still has to apply every current validation rule. Your algorithm already keeps persistUntil, MPIC, and CAA on that path; policy and any parameter the CA has since started recognizing belong on the list too. issuer-domain-name can't be skipped either as BR 3.2.2.4.22 requires it to be a name the CA currently discloses in its CP/CPS, and embedding it in the stored text doesn't prove the CA still accepts that identity. The specification would also need to define "exact": raw TXT RDATA preserves character-string boundaries that do not affect the parsed issue-value, while a parsed comparison needs canonicalization.

sheurich: > Im TOTALLY against a extra request for "verifying" a accounturi. Why?
> 
> Because, to be effective, the check must be made mandatory - for example by adding a token value that must be inserted into DNS. (which defeats the purpose of a "PERSIST")
> 
> And it also wastes resources by additional requests.

The account-resource proof doesn't require a token in DNS. It is a mandatory client construction check before publication: sign a POST-as-GET to the exact account URL used in the digest, following no redirects, and require the response to prove that the key controls that account resource. A conforming client refuses to publish if the check fails. The vector it closes exists inside client software because that is the code computing the digest from a URL it was handed, so the check sits where the threat does; a DNS token would turn a persistent method into an interactive one. The open caching proposal is one proof per unchanged (account URL, account key) pair, not one request per order or record. It targets the honest-client foreign-URL path only. Pasted complete values, direct DNS writes, and a TLS-terminating proxy remain residuals.

sheurich: > Note that the mechanism (with disabling "concurrent" deactivated keys cross-account) you are describing would not prevent the H(domain, victim key, attacker URL) attack, since the victim key does not belong to a deactivated account. The attacker has control of the victim key as soon as it leaks.
> 
> Only way this would be prevented is blacklisting old keys after each keychange, so keyChanged keys effectively becomes attached to a "new virtual" account that is automatically deactivated.

Agreed that neither account deactivation nor the RFC 8555 7.3.5 duplicate-key check prevents H(domain, victim key, attacker URL); my comment cited the duplicate-key check to explain how the planted record matures, not as a defense against it. A CA-side check helps only when the target CA previously saw the key, and the form matters. Refusing a keyChange that would bind a key the CA witnessed in another account's history closes the same-CA variant without touching legitimate accounts. Treating every rotated-out key as deactivated does not work: it would stop records provisioned under prior keys from validating for their own account, which is the non-breaking rollover property. In @MaxHearnden's variant where the victim never had an account at the target CA, the CA has no key history, so no check engages; the leak alone lets the attacker bind that key to the pre-registered account. A keyChange check can complement the client-side account-resource proof but cannot replace it.

sebastiannielsen: The problem with the check is that its not technically mandatory. Meaning, if the client does not publish the record itself, it could still be copy+pasted into the record and thus bypass the check.

And as I said, it creates an extra round-trip to the server which I think is bad, and it requires every hashed account-URI to be resolvable.

Another thing is that the check you are describing, does not prevent a integrator flow or similiar from displaying an attacker's value. Because then theres no ACME client on the client side (which visit the webpage) that can do the check in a thrustworthy way.


My proposed solution of requiring the whole RDATA to be stored along with account ID and domain is the best solution here.

When I say the CA does not need to check "issuer-domain-name" is because that is already checked in the string comparision check of the record. A changed issuer-domain-name would make the record not match the stored value from the beginning, **thus no additional issuer-domain-name check is required**.
This is however required when doing the check where the values are recomputed to the hash.


No canonalization is required. It should be stored bit for bit as-is.
If any spacing or similiar is changed, the record does no longer match the stored value, and thus the record would become invalid for rotated-key scenarios.

For MPIC purposes, if multiple DNS servers would return a different result while the record still conveys the same intent, the request from the CA's main datacenter is the one that gets stored and what gets compared to, which means as long as the domain owner's primary and secondary DNS servers return the record in the exact same format, exact same spacing and exact same captilization, the record will be accepted for rotated key scenarios.

sheurich: The proposed account-resource check does not address pasted values or nonconforming integrator flows; those are residual risks I already acknowledged. Whole-RDATA storage does not close them either: a planted value that passes first validation becomes the stored value under either design. The check does address the honest-client path, where software computing the digest is handed a foreign account URL.

It is an extra request, but it targets the ACME account URL, not the hashed `accounturi`, so nothing requires a hashed URI to resolve. RFC 8555 Section 7.3 defines that URL as the account resource and the target for account-management requests. The proposed cache scope is one proof per unchanged (account URL, account key) pair, not one request per order or record.

`issuer-domain-name` still needs current evaluation. BR 3.2.2.4.22 requires the name to be disclosed in Section 4.2 of the CA's CP/CPS at the time of each validation, and the draft requires it to appear in the current challenge's `issuerDomainNames`. Stored text proves only what the CA accepted earlier.

A bitwise design also has to define the compared representation: raw DNS wire RDATA, concatenated TXT character-strings, or parsed `issue-value`. Raw equality makes segmentation significant, so one character-string and two carrying the same `issue-value` do not match, and RFC 8659 allows optional whitespace inside `issue-value`. Both the draft and BR 3.2.2.4.22 require unknown parameters to be ignored, so under raw equality adding an inert parameter invalidates a record after rollover. The design also needs to say which RR of a multi-record, multi-issuer RRset is the stored one, and what a later RRset addition or removal does to the match.

MPIC adds a further constraint on any whole-RDATA equality rule. BR 3.2.2.9 forbids reusing or caching one perspective's result for another, so each perspective used has to retrieve the record itself and the CA has to assess those observations against the stored representation; the quorum table allows one non-corroboration for two to five remote perspectives and two for six or more. If only the primary runs the historical equality check, the remote perspectives never corroborate the determination that gates issuance. If every observation is checked, raw equality is stricter than the same-`accounturi` corroboration BR 3.2.2.4.22 requires, and one secondary that segments the same value differently consumes the entire allowance in a deployment with two to five remote perspectives.

sebastiannielsen: **1: If we add a accountURI check ( according to pull #62 ) , we don't really need the whole accounthash construction.**
The whole idea with requiring client-computed information is just to prevent a hijacked or malicious ACME server from issuing a attacker's accountURI.

**2: When I said issuer-domain-name didn't need current evaluation, I meant this:**

CA checks issuer-domain-name, MPIC, persistUntil, and then composes a hashed AccountURI using Accounts current JWK hash, * or domain, and the accountURI.
Now CA finds these values valid and marks autz as valid.

CA now stores the record: "letsencrypt.org; accounturi=http://acme.invalid/hashprefix/g6Y_gf235tfd-djdue_sue; policy=wildcard" in database along with Domain (where the record lives, CAN include _validation-persist if CA wishes), and the account that performed the autz.

Account performs a keyChange

Now 80 (or 40 days) the domain owner now does a renew, keeping the old record (which don't match the key).

Now CA looks in the database:
"does the record "letsencrypt.org; accounturi=http://acme.invalid/hashprefix/g6Y_gf235tfd-djdue_sue; policy=wildcard" exist in the database for the right accounturl and database.

YES we found the record.

Now the CA *ONLY* needs to check persistuntil and MPIC and policy. They do not need to check the "issuer-domain-name" because according to the standard, **no old "issuer-domain-name" can go invalid.**
If letsencrypt.org was accepted previously, it will ALWAYS be accepted.

**And this was already checked by looking up the value in the database.**
If the record would say instead:
"letsencrypt_v3.org; accounturi=http://acme.invalid/hashprefix/g6Y_gf235tfd-djdue_sue; policy=wildcard"

the value would be REJECTED, since it doesn't exist in database, even if "letsencrypt_v3.org" would be an accepted issuer-domain-name. It would need to be combined with a current key of the account to become accepted.

So the issuer-domain-name is already checked, because thats PART of the value being looked up. So the check "**check if letsencrypt.org is on the list of issuer-domain-names we accept**" does not need to be performed. It have already been checked once.

This gives another advantage:
**The CA does NOT need to store old issuer-domain-names they previously used in their acceptance list. Once a issuer-domain-name is scrapped, it can be removed from their list - because the stored-record system will ensure old records still validate even if they bear a issuer-domain-name no longer in use.**

**3: Here is the imporant piece. When doing caonicalization, it sohuld be checked as "concatenated TXT strings", unfolded.**
Meaning whitespace should be included:
`"letsencrypt_v3.org; accounturi=http://acme.invalid/hashprefix/g6Y_gf235tfd-djdue_sue; policy=wildcard"`
and
`"letsencrypt_v3.org;accounturi=http://acme.invalid/hashprefix/g6Y_gf235tfd-djdue_sue;policy=wildcard"`
is not the same thing. This is a CHANGED record and needs a new key.

Note that multiple records under the same label should be counted separately. Meaning adding or removing other records should NOT affect this type of validation. This allows a DNS operator to add a new DNS-PERSIST-01 record belongning to a new key, while keeping a old record, to facilitate key rollover.

**4: About MPIC validation, its when the database is looked up, then the values should be checked by the main CA datacenter.**

So if sub-MPIC datacenters return a slightly different (but still equal records when checking issuer-domain-name, accounturi and policy) for example due to different recursive resolver software, then MPIC should still consider these records corroborating.

But the corroboration datacenter should never make the initial CA stored-value database lookup, because then these small differences may cause a rotated key record to fail validation randomly (depending on which MPIC datacenter who does the database lookup).

Thats why its important its the same datacenter that did the lookup when the key was current - that must do the lookup now when the key have rotated and the CA has to rely on a stored record.

This means, the main CA server, which rely on a stored-value-lookup, will still parse the values before sending the parsed values to the sub-MPIC servers.

So the sub-MPIC servers will still check for corroborating records the "normal" way and not require an exact match. The exact match to the database is ONLY needed for the main CA server that does the "initial" lookup.

sheurich: Taking your four points in order.

1. The account check and the hash do different jobs, so the first cannot replace the second. Whichever variant you mean, #62's token or the plain authenticated POST-as-GET I proposed, the check runs in the client, and the CA cannot verify it happened. The hash is the part the CA enforces: it recomputes the value over the requesting account's URL and that account's keys. With a plain account URL in the record, a client handed a foreign URL publishes a record that validates for the foreign account immediately, with no key compromise. With the hashed form, the same plant validates only if the attacker's account later holds the victim's key through `keyChange`, and RFC 8555 Section 7.3.5 requires the inner JWS to be signed by that new key, so the attacker needs the victim's private key. The hash also keeps the account URL out of public DNS, and the domain input keeps one account's records at different domains uncorrelated; both are stated goals of the record format, not side effects. The draft dropped the plain form deliberately.

2. BR 3.2.2.4.22 does not support "no old issuer-domain-name can go invalid." Its RDATA requirements are things the CA "MUST confirm" at each validation, and the second is that the `issuer-domain-name` value "MUST be an Issuer Domain Name disclosed by the CA in Section 4.2 of the CA's Certificate Policy and/or Certification Practices Statement." Present tense, no stored-match exemption, and this method's 10-day reuse cap makes every later acceptance a fresh validation. The advantage you describe, retiring a name from your acceptance list while stored records bearing it keep validating, is the issuance that requirement 2 forbids. Your own model concedes the underlying point: `policy` (and `persistUntil` when present) is also part of the looked-up value, yet you evaluate both fresh. Database equality preserves whatever the CA accepted earlier; it is not a currency check.

3. Concatenated character-strings and per-RR matching answer two of my questions: segmentation stops mattering, and a record for a new key can sit beside the old one. Two problems remain. RFC 8659 Section 4.2 permits whitespace around the `issuer-domain-name`, the semicolons, and parameter tags, and assigns it no meaning, so under whitespace-exact matching a provider that reformats a zone cosmetically turns a still-conformant record into a "changed" one needing current-key revalidation after rollover. And BR requirement 5 obliges CAs to ignore unknown parameter keys, so adding an inert parameter still kills a stored match. The draft already requires Simple String Comparison for the `accounturi` value itself (RFC 3986 Section 6.2.1): exactness where it carries security.

4. On structure we agree: perspectives never make the determination, and method 22's corroboration standard, a record demonstrating control with the same `accounturi` as the primary, is semantic, so your perspective-side rule satisfies it. I withdraw the quorum-consumption objection for this variant; what remains is that the equality gate runs from one vantage. The premise still needs correcting: recursive resolvers return TXT RDATA as the zone serves it, segmentation included. If perspectives see different bytes, the zone is serving different data, which is what MPIC exists to surface. Deterministic concatenation yields the same comparison at every datacenter, so pinning stored-value lookups to the datacenter that saw the first validation has no BR basis and ties record validity to CA topology: decommission that datacenter and its records strand.

None of this has landed as code here, so the review blockers stand.

sebastiannielsen: 2: aha, I understand now. Ill add that to my list that issuer-domain-name needs to be checked.
I tought the DNS-PERSIST-01 would require the CA to also accept old values, why this uses a separate issuer-domain-names instead of just using caaIdentities.

3: Correct, adding a inert parameter kills the record. But it doesn't kill it permanently, you can remove the inert parameter to restore the record.
The idea is just that any changes to the record makes it no longer a historic record that needs preservation, thus it must be made with a current key of the account.

4: Exactly. It doesn't matter if a datacenter is decommisioned, because the lookup will be made from another "main" datacenter.
What I wanted to guard against, is that a CA uses software, that returns the MPIC results in a slightly different format (maybe in different character encodings, a slightly different format and so on) which may cause validation to fail at an exact match, and thus lead to intermittient issues like authorization for keyChanged keys **randomly** failing.
Thats why I wanted to make sure, that the "main check" against the stored record is only done from one vantage point, and at other vantage points, the check is done as "usual" (like the record would have been hashed using a current key).

**Once the stored record matches, everything from that point on, should be treated like the record was made using a current account key.**


And you are correct that none of this have landed as code, Im just colleting all informations, and as soon as I have all information I need, I will update the pull to match the review blockers.

